PR #1, opened 2020-06-11
Move Swift code into the root

AquaGeek:
This moves all the stuff that used to live in `/swift` to the root of the repo. It also removes some files that were needed for generating documentation (which now lives in the original repo).


README.md
```diff
@@ -1,5 +1,6 @@
 # workflow
 
+![Swift CI](https://github.com/square/workflow/workflows/Swift%20CI/badge.svg)
```
dhavalshreyas, line 3: Maybe? 

```suggestion
![Swift CI](https://github.com/square/workflow-swift/workflows/Swift%20CI/badge.svg)
```

PR #3, opened 2020-06-12
Update ReactiveSwift

dhavalshreyas:
`ReactiveSwift 6.0.0` was released more than a year ago now + there were Xcode warnings that will now be fixed.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.2
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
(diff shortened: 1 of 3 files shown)


AquaGeek:
Should we bump `Package.swift` as well?

dhavalshreyas:
I think we should, however for some reason I'm bumping into:
```
error: the package dependency graph could not be resolved; unable to find any available tag for the following requirements:
    https://github.com/ReactiveCocoa/ReactiveSwift.git @ 6.3.0..<7.0.0
```

dhavalshreyas:
Looks like `ReactiveSwift` `6.0.0` requires `swift-tools-version:5.2`. Not sure if this will trip up our CI. 🤞🏽

JustinDSN:
Uh oh, which version of XCode comes with Swift-Tools version 5.2? Do we have it with XCode 11.2.1?

dhavalshreyas:
> Uh oh, which version of XCode comes with Swift-Tools version 5.2? Do we have it with XCode 11.2.1?

Nope, Xcode 11.4.

PR #7, opened 2020-06-17
Allow updating the root workflow in a ContainerViewController

bencochran:
Especially when transitioning things to Workflow, when you have many small root workflows, it can be useful to feed in new properties to the root. Right now, most things I’ve seen that need to update this root workflow will initialize it with a `Property<T>` of something that can be changed from the outside later. A nicer pattern is to embrace the declarative nature of Workflow itself and allow it to be updated when necessary.

We originally worked on this and added `update(workflow:)` to `WorkflowHost`, but the fact that it was held as an `Any` in `ContainerViewController` prevented us from getting all the way there.

To accomplish this here, we now always wrap the root workflow. This allows us to keep a typed `WorkflowHost` instance that we can later update. A side benefit is we no longer need an extension to host an `AnyWorkflowConvertible`, and it can be the primary initializer now.

One thing to note is that I could initialize a `ContainerViewController` with `WorkflowA` and then update it with `WorkflowB` so long as those two had the same `Output` and `Rendering`. I think this is actually a feature, since this mimics how you can switch from rendering one workflow to another _inside_ a workflow seamlessly.

## Checklist

- [X] Unit Tests


zach-klippenstein:
I was reluctant about this at first, but then I remembered Swift doesn't have a props type so this is already functionality we have in Kotlin, in spirit, I think (see https://github.com/square/workflow/pull/1180).

bencochran:
@zach-klippenstein Yeah, this largely got re-motivated in my head seeing that discussion go by (or, from what I was able to follow of it).

bencochran:
CLA bot looks stuck on this one…

dhavalshreyas:
Done.

PR #11, opened 2020-06-26
WorkflowTesting changes to support GUWT

dhavalshreyas:
The new `SignalProducerWorker` will use a `SignalProducerWorkerWorkflow` to run `SignalProducer`. 

To be able to build `ExpectedSignalProducerWorker`, `ExpectedWorkflow` will need to support custom matching (b/w the expected and the actual `Workflow`) and assertions.

Changes in this PR will enable this.

### DIFF
--- a/WorkflowTesting/Sources/RenderExpectations.swift
+++ b/WorkflowTesting/Sources/RenderExpectations.swift
@@ -15,6 +15,7 @@
  */
 
 import Workflow
+import XCTest
 
 /// A set of expectations for use with the `WorkflowRenderTester`. All of the expectations must be fulfilled
 /// for a `render` test to pass.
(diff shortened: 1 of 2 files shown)


WorkflowTesting/Sources/RenderExpectations.swift
```diff
@@ -123,15 +124,44 @@ extension ExpectedSideEffect {
...
 public struct ExpectedWorkflow {
-    let workflowType: Any.Type
-    let key: String
-    let rendering: Any
-    let output: Any?
-
-    public init<WorkflowType: Workflow>(type: WorkflowType.Type, key: String = "", rendering: WorkflowType.Rendering, output: WorkflowType.Output? = nil) {
-        self.workflowType = type
-        self.key = key
+    public let rendering: Any
```
bencochran, line 127: Exposing the type-erased internals here worries me. Do these need to be public?

dhavalshreyas: Actually, I don't think so. Dropping `public` here.

WorkflowTesting/Sources/RenderExpectations.swift
```diff
@@ -123,15 +124,44 @@ extension ExpectedSideEffect {
 }
 
 public struct ExpectedWorkflow {
-    let workflowType: Any.Type
-    let key: String
     let rendering: Any
     let output: Any?
+    let doesMatch: (_ childWorkflow: Any, _ childKey: String) -> Bool
```
AquaGeek, line 129: This sounds strange to me. Can we call this `matches` instead?

dhavalshreyas: Closing this one, in favor of #15

dhavalshreyas:
Closing this one in favor of #15

PR #12, opened 2020-06-26
Move Worker out of Workflow library to WorkflowReactiveSwift

dhavalshreyas:
Moves `Worker` to a different library `WorkflowReactiveSwift`. The new `Worker` is built on top of the `SideEffect` primitive.


Workflow/Tests/ConcurrencyTests.swift
```diff
@@ -319,7 +289,8 @@ final class ConcurrencyTests: XCTestCase {
         signal1.send(value: 1)
         signal2.send(value: 2)
 
-        XCTAssertEqual(0, host.rendering.value.count)
+//        TODO: What does it mean that this is no longer true?
+//        XCTAssertEqual(0, host.rendering.value.count)
```
dhavalshreyas, line 293: @bencochran, @AquaGeek : I think this is okay, right? When we do a `signal.send()` we synchronously update the `rendering`, unlike with the `Worker`s. Just wanted to confirm before removing this.

Workflow/Sources/RenderContext.swift
```diff
@@ -71,10 +71,6 @@ public class RenderContext<WorkflowType: Workflow>: RenderContextType {
         fatalError()
     }
 
-    public func awaitResult<W, Action>(for worker: W, outputMap: @escaping (W.Output) -> Action) where W: Worker, Action: WorkflowAction, WorkflowType == Action.WorkflowType {
```
AquaGeek, line 74: Are we not providing a backwards-compatible deprecated shim for these?

dhavalshreyas: I was thinking not, but I'll go ahead and add it in. We'll still need to do the `import WorkflowReactiveSwift`, but I'll expose the same APIs there.

Workflow/Tests/SubtreeManagerTests.swift
```diff
@@ -140,163 +140,6 @@ final class SubtreeManagerTests: XCTestCase {
...
-                state: .notWorking,
-                context: context
-            )
-        }
-
-        XCTAssertFalse(isStillRunning)
-        wait(for: [endExpectation], timeout: 1.0)
-    }
-
-    func test_subscriptionsUnsubscribe() {
```
bencochran, line 231: We should test this still too in `WorkflowReactiveSwitftTests`

Workflow/Tests/WorkflowNodeTests.swift
```diff
@@ -195,63 +195,6 @@ final class WorkflowNodeTests: XCTestCase {
 
         XCTAssertEqual(snapshot, expectedSnapshot)
     }
-
-    func test_handlesRepeatedWorkerOutputs() {
```
bencochran, line 199: This to `WorkflowReactiveSwitftTests` too please

Workflow/Tests/ConcurrencyTests.swift
bencochran: The `.observe(on: QueueScheduler.workflowExecution)` when subscribing to the signal producer inside `ChildWorker` before meant that all worker outputs were received asynchronously on the next run loop cycle. This exposes that behavior change. I’m not entirely convinced that was intentional behavior or an unintended side-effect that got codified into this test though.

WorkflowTesting/Tests/WorkflowRenderTesterDeprecatedTests.swift
```diff
@@ -107,63 +107,11 @@ final class WorkflowRenderTesterDeprecatedTests: XCTestCase {
             )
     }
 
-    func test_workers() {
```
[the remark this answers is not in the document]
dhavalshreyas: I was hoping we can just use `WorkflowReactiveSwiftTesting` for the `WorkflowReactiveSwiftTests` and call it done. :D 

But yup, splitting them is the right thing to do.

Workflow/Tests/SubtreeManagerTests.swift
dhavalshreyas: I wasn't sure if this was left over from the old `Subscriptions` world.

dhavalshreyas: I'm not sure if we're testing anything relevant here. Dropping this one unless you believe it does add value.

Workflow/Sources/RenderContext.swift
dhavalshreyas: Done.

WorkflowReactiveSwift/Sources/SignalProducerWorkflow.swift
```diff
@@ -0,0 +1,60 @@
...
+    public typealias Output = Value
+    public typealias State = Void
+    public typealias Rendering = Void
+
+    var signalProducer: SignalProducer<Value, Never>
+    var key: AnyHashable
+
+    public init(signalProducer: SignalProducer<Value, Never>, key: AnyHashable = "") {
+        self.signalProducer = signalProducer
+        self.key = key
```
dhavalshreyas, line 37: I don't think this `key` makes any sense, right? There'll always be just one `sideEffect` in this workflow. cc: @bencochran

bencochran: Yeah, agreed. Consumers can provide a key when running this workflow itself.

WorkflowReactiveSwift/Sources/Worker.swift
```diff
@@ -24,19 +26,51 @@ import ReactiveSwift
...
+    }
+
+    func workflowDidChange(from previousWorkflow: WorkerWorkflow<WorkerType>, state: inout UUID) {
+        if !worker.isEquivalent(to: previousWorkflow.worker) {
+            state = UUID()
+        }
+    }
+
+    func render(state: State, context: RenderContext<WorkerWorkflow>) -> Rendering {
+        // Start with Void to ensure `worker.run()` is called only once for a given key
```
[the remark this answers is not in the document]
dhavalshreyas: Not sure if I like the alternatives better. If we do the `runSideEffect` directly - we'll need to repeat the `sink`, `lifetime` management etc.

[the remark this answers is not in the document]
bencochran: From my perspective that stuff’s easier to glance at and understand, whereas this I had to stare at for a second to get _why_ it had that effect. I’m not hardline on it, but I think the subtlety is tricky.

WorkflowReactiveSwift/TestingTests/TestingTests.swift
```diff
@@ -0,0 +1,197 @@
...
+            )
+
+        expectingFailures([
+            #"Workers of type TestWorker not equivalent. Expected: TestWorker(input: "not-test"). Got: TestWorker(input: "test")"#,
+        ]) {
+            tester.render { _ in }
+        }
+    }
+
+//    TODO(dhaval): This currently fatals because it's not sure what Rendering to return.
```
bencochran, line 75: Ah, right. We can’t actually not expect Workflows and pass tests. Seems fine to remove this.

WorkflowReactiveSwiftTesting.podspec
```diff
@@ -0,0 +1,25 @@
...
+
+  s.source_files = 'WorkflowReactiveSwift/Testing/**/*.swift'
+
+  s.dependency 'Workflow', "#{s.version}"
+  s.dependency 'WorkflowReactiveSwift', "#{s.version}"
+  s.dependency 'WorkflowTesting', "#{s.version}"
+  s.dependency 'ReactiveSwift', '~> 6.3.0'
+
+  s.framework = 'XCTest'
+end
```
bencochran, line 25: nit: We should probably define the test_spec for the associated tests here

Workflow/Tests/SubtreeManagerTests.swift
bencochran: Yeah, I guess it currently ends up being a test of SignalWorker, which should have its own independent test coverage.

WorkflowReactiveSwift/Sources/Worker.swift
```diff
@@ -24,19 +26,51 @@ import ReactiveSwift
 /// If there is, and if the workers are 'equivalent', the context leaves the existing worker running.
 ///
 /// If there is not an existing worker of this type, the context will kick off the new worker (via `run`).
-public protocol Worker {
+public protocol Worker: AnyWorkflowConvertible {
```
AquaGeek, line 29: Nit: Should the conformance be moved to the extension below?

Workflow/Sources/WorkflowLogger.swift
```diff
@@ -88,41 +88,4 @@ final class WorkflowLogger {
             os_signpost(.end, log: .workflow, name: "Render", signpostID: signpostID)
         }
     }
-
-    // MARK: - Workers
-
-    static func logWorkerStartedRunning<WorkerType>(ref: AnyObject, workerType: WorkerType.Type) {
```
AquaGeek, line 94: Should these be moved to the other pod?

If we decide to remove them altogether we should also remove the definition of the `worker` log handle, at the top.

AquaGeek:
Have we verified that `ReactiveSwift`- and `RxSwift`-based workers play nicely together/can coexist in the same workflow?

Workflow/Sources/WorkflowLogger.swift
dhavalshreyas: Since `Workers` are `Workflows` now, we should be able to get all the logs through those. 

I did notice we don't logs for `SideEffects` though.

Development.podspec
```diff
@@ -11,6 +11,7 @@ Pod::Spec.new do |s|
   s.swift_version = '5.0'
   s.dependency 'Workflow'
   s.dependency 'WorkflowUI'
+  s.dependency 'WorkflowReactiveSwift'
```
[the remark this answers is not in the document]
dhavalshreyas: Yes, looking to see if I've missed adding it elsewhere.

WorkflowReactiveSwift/Sources/Worker.swift
dhavalshreyas: Since `Worker` is a protocol, we get the `Extension of a protocol cannot have an inheritance clause` error.

dhavalshreyas:
> Have we verified that `ReactiveSwift`- and `RxSwift`-based workers play nicely together/can coexist in the same workflow?

Not yet, I'll test this when we do the `WorkflowRxSwift`

Workflow/Sources/WorkflowLogger.swift
bencochran: Oh yeah, it would be nice to log the start and end of side-effects for sure

Development.podspec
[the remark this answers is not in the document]
dhavalshreyas: I checked, I think we're good.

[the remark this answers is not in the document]
AquaGeek: (Sorry, I was referring to internal pods. i.e. This will need to be done as part of the upgrade to 1.0ɑ.)

WorkflowReactiveSwift/Sources/Worker.swift
AquaGeek: Ah. Yes. 👍

Development.podspec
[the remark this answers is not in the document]
dhavalshreyas: Ahh yes! That and adding `import WorkflowReactiveSwift` will be the two required migration steps.

Workflow/Sources/WorkflowLogger.swift
AquaGeek: We'll have to figure out how to filter `Worker` logs from the `Workflow` ones, so we don't end up double-logging.

dhavalshreyas:
🙏 I'm adding the `WorkerLogs` in a follow-up PR.

PR #15, opened 2020-06-28
Introduce new RenderTester API

bencochran:
Per #16 and https://github.com/square/workflow/issues/702, introduce a new RenderTester API shaped like the following:

```swift
NameLoadingWorkflow()
    .renderTester(initialState: .init(state: .loading, token: "user-token"))
    .expect(
        worker: LoadingWorker(token: "user-token"),
        producingOutput: .success("Ben Cochran")
    )
    .expectSideEffect(key: "some-extra-side-effect-just-for-fun")
    .render { rendering in 
        XCTAssertEqual(rendering.title, "Loading")
    }
    .verify(
        action: MyWorkflow.Action.loadSuccess(name: "Ben Cochran")
    )
    .verify(
        output: .complete
    )
```

The `expect` methods come in the following shapes:

* **Workflows**:
    * `expectWorkflow(workflowType: WorkflowType.Type, key: String, producingRendering: WorkflowType.Rendering, producingOutput: WorkflowType.Output?, assertions: ((WorkflowType) -> Void)?)`
* **Workers**
    * `expect(worker: WorkerType, producingOutput: WorkerType.Output?)`
* **Side effects**
    * `expectSideEffect(key: AnyHashable, producingAction: ActionType?)`

The `verify` methods come in the following:

* **Actions**
    * `verifyAction(assertions: (Action) -> Void)`
    * `verify(action: Action)` (when Equatable)
    * `verifyNoAction()`
* **Output**
    * `verifyOutput(assertions: (Output) -> Void)`
    * `verify(output: Output)` (when Equatable)
    * `verifyNoOutput()`
* **Resulting state**
    * `verifyState(assertions: (State) -> Void)`
    * `verify(state: State)` (when Equatable)

This also deprecates, but keeps around, the old API (you’ll see some type erasure and contortions in this PR to reimplement the old API on top of the new one).

## Checklist

- [X] Unit Tests
- [ ] ~UI Tests~
- [ ] ~Snapshot Tests (iOS only)~


zach-klippenstein:
I like the equatable overloads for state and output, filed for Kotlin as https://github.com/square/workflow-kotlin/issues/63

dhavalshreyas:
This is really cool! I was worried about having to overload the same `RenderExpectations` for the different types of `Worker`s. This will help a lot.

WorkflowTesting/Sources/Internal/RenderExpectations.swift
```diff
@@ -0,0 +1,91 @@
...
+                self.workflowType = workflowType
+                self.key = key
+            }
+        }
+
+        internal class ExpectedWorkflow<ExpectedWorkflowType: Workflow>: AnyExpectedWorkflow {
+            let rendering: ExpectedWorkflowType.Rendering
+            let output: ExpectedWorkflowType.Output?
+
+            init(key: String, rendering: ExpectedWorkflowType.Rendering, output: ExpectedWorkflowType.Output?) {
```
dhavalshreyas, line 35: I will need something like in: https://github.com/square/workflow-swift/pull/11/files#diff-d274c3ab6b1b700864fdd0b96d30b731R129

to support custom matching for `Worker`s.

bencochran: Do we? We don’t actually want to reproduce/change the matching behavior since the RenderContext won’t have additional matching logic for Workers. I’d pictured something like this:
```swift
extension RenderTester {
    public func expect<ExpectedWorkerType: Worker>(
        worker expectedWorker: ExpectedWorkerType,
        key: String = "",
        producingOutput output: ExpectedWorkerType.Output? = nil
    ) -> RenderTester<WorkflowType> {
        return self.expectWorkflow(
            type: WorkerWorkflow<ExpectedWorkerType>,
            key: key,
            producingRendering: (),
            producingOutput: output,
            assertions: { actualWorker in
                XCTAssert(expectedWorker.isEquivalent(to: actualWorker))
            }
        )
    }
}
```

(though I _do_ think I should update `assertions` to give the file and line so we can correctly place the error where it should be)

dhavalshreyas: Ooh, that's right, fantastic!

A slight change, we'll get the `Workflow` here.

```swift
extension RenderTester {
    public func expect<ExpectedWorkerType: Worker>(
        worker expectedWorker: ExpectedWorkerType,
        key: String = "",
        producingOutput output: ExpectedWorkerType.Output? = nil
    ) -> RenderTester<WorkflowType> {
        return self.expectWorkflow(
            type: WorkerWorkflow<ExpectedWorkerType>,
            key: key,
            producingRendering: (),
            producingOutput: output,
            assertions: { actualWorkflow in
                XCTAssert(actualWorkflow.worker.isEquivalent(to: expectedWorker))
            }
        )
    }
}
```

bencochran: Ah, right, yep!

bencochran:
After noting “though I _do_ think I should update `assertions` to give the file and line so we can correctly place the error where it should be” I realized it would be nicer if expectations asserted on the line of the expectation rather than on `render`, so I updated to do that (which also means `assertions` on `expectWorkflow` doesn’t need file and line ✨).

I also fixed the fact that `WorkflowRenderTesterFailureTests` wouldn’t compile on Xcode 11 and filed #18.

WorkflowTesting/Sources/Internal/AppliedAction.swift
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Workflow
+import XCTest
+
+struct AppliedAction<WorkflowType: Workflow> {
```
AquaGeek, line 20: Does this need to be wrapped in an `#if DEBUG`?

bencochran: I don’t _think_ so if the comment about `@testable import` being the issue was correct (since this file does a standard import)

PR #26, opened 2020-07-02
Add os_log logging to Workers

dhavalshreyas:
Follow-up from #12, we had dropped the logging there, adding it back in here.

### DIFF
--- a/WorkflowReactiveSwift/Sources/Logger.swift
+++ b/WorkflowReactiveSwift/Sources/Logger.swift
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import os.signpost
+
+private extension OSLog {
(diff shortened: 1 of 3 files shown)


WorkflowReactiveSwift/Sources/Worker.swift
```diff
@@ -63,14 +63,33 @@ struct WorkerWorkflow<WorkerType: Worker>: Workflow {
     }
 
     func render(state: State, context: RenderContext<WorkerWorkflow>) -> Rendering {
+        let logger = WorkerLogger<WorkerType>()
+
         // Start with Void to ensure `worker.run()` is called only once for a given key
         SignalProducer(value: ())
-            .flatMap(.latest) { self.worker.run() }
+            .flatMap(.latest) {
+                logger.logStarted()
```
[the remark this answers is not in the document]
bencochran: Like maybe an `on(started:)` hanging off the `.run()`?

WorkflowReactiveSwift/Sources/Logger.swift
```diff
@@ -0,0 +1,65 @@
...
+    // MARK: - Workers
+
+    func logStarted() {
+        if #available(iOS 12.0, macOS 10.14, *) {
+            os_signpost(
+                .begin,
+                log: .worker,
+                name: "Running",
+                signpostID: self.signpostID,
+                "Worker: %{public}@",
```
bencochran, line 41: Should we consider these non-public? If I were implementing a Worker to do login, I’d likely put a username & password into the properties of that Worker.

AquaGeek: We're logging the type of the worker, not the worker itself.

Realistically, we can probably move to making all of these non-public, though.

dhavalshreyas: Cool, will go ahead and make these private?

dhavalshreyas: Done.

bencochran: Oops, totally missed that it was the type not the value.

PR #27, opened 2020-07-02
Introduce WorkflowRxSwift

dhavalshreyas:


WorkflowRxSwift/Tests/WorkerTests.swift
```diff
@@ -0,0 +1,215 @@
...
+ * limitations under the License.
+ */
+
+import Foundation
+import RxSwift
+import WorkflowRxSwift
+import WorkflowRxSwiftTesting
+import WorkflowTesting
+import XCTest
+@testable import Workflow
```
bencochran, line 23: I’d avoid `@testable` importing a module other than the one under test (i.e. treat this module as a third-party module that can’t access internals of Workflow). Can we use `WorkflowTesting` to help us here?

PR #28, opened 2020-07-02
Recover from unexpected void-rendering workflows in RenderTester

bencochran:
`Void` only has a single value, so it’s reasonable/possible for `RenderTester`’s `RenderContext` to return a rendering for unexpected workflows that render `Void`.

Notably, this allows us to fail but continue testing when we encounter an unexpected `Worker` (and thus we can bring back `test_worker_unexpected`).

### DIFF
--- a/WorkflowReactiveSwift/TestingTests/TestingTests.swift
+++ b/WorkflowReactiveSwift/TestingTests/TestingTests.swift
@@ -72,6 +72,15 @@ class WorkflowReactiveSwiftTestingTests: XCTestCase {
         }
     }
 
+    func test_worker_unexpected() {
+        let tester = TestWorkflow()
+            .renderTester(initialState: .init(mode: .worker(input: "test"), output: ""))
+
+        expectingFailure(#"Unexpected workflow of type WorkerWorkflow<TestWorker> with key """#) {
+            tester.render { _ in }
+        }
+    }
+
     // MARK: - Failure Recording
 
     var expectedFailureStrings: [String] = []
--- a/WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
+++ b/WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
@@ -63,6 +63,12 @@
                         diagnosticMessage = ""
                     }
                     XCTFail("Unexpected workflow of type \(Child.self) with key \"\(key)\". \(diagnosticMessage)", file: file, line: line)
+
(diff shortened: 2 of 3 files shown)


WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
```diff
@@ -114,4 +120,26 @@
         }
     }
 
+    func defaultRendering<Child>(for workflowType: Child.Type) -> Child.Rendering? where Child: Workflow {
```
dhavalshreyas, line 123: Are these used?

bencochran: Oh whoops, these were my failed attempts at avoiding the `() as! Child.Rendering`. Removing.

WorkflowReactiveSwift/TestingTests/TestingTests.swift
```diff
@@ -72,6 +72,15 @@ class WorkflowReactiveSwiftTestingTests: XCTestCase {
         }
     }
 
+    func test_worker_unexpected() {
+        let tester = TestWorkflow()
+            .renderTester(initialState: .init(mode: .worker(input: "test"), output: ""))
+
+        expectingFailure(#"Unexpected workflow of type WorkerWorkflow<TestWorker> with key """#) {
```
AquaGeek, line 79: Wait, what kind of wizardry is this?

WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
```diff
@@ -63,6 +63,12 @@
                         diagnosticMessage = ""
                     }
                     XCTFail("Unexpected workflow of type \(Child.self) with key \"\(key)\". \(diagnosticMessage)", file: file, line: line)
+
+                    // We can “recover” from missing Void-rendering workflows since there’ss only one possible value to return
```
AquaGeek, line 67: Typo: too many "s"sssss

WorkflowReactiveSwift/TestingTests/TestingTests.swift
bencochran: [SE-200](https://github.com/apple/swift-evolution/blob/master/proposals/0200-raw-string-escaping.md)!

WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
bencochran: 😅

PR #30, opened 2020-07-02
Do not use @testable import Workflow in WorkflowReactiveSwiftTests

dhavalshreyas:
Fixing @bencochran 's feedback on `WorkflowRxSwiftTests` for `WorkflowReactiveSwiftTests`: https://github.com/square/workflow-swift/pull/27#discussion_r449185026

### DIFF
--- a/WorkflowReactiveSwift/Tests/WorkerTests.swift
+++ b/WorkflowReactiveSwift/Tests/WorkerTests.swift
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import Foundation
 import ReactiveSwift
 import Workflow
 import WorkflowReactiveSwift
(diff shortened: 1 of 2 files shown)


WorkflowReactiveSwift/Tests/WorkerTests.swift
```diff
@@ -64,23 +64,26 @@ class WorkerTests: XCTestCase {
...
+            let initialState: State
+
+            func makeInitialState() -> State {
+                initialState
             }
 
-            func makeInitialState() -> WorkerWorkflow.State {
-                return .notWorking
+            func workflowDidChange(from previousWorkflow: WorkerWorkflow, state: inout State) {
+                state = initialState
```
bencochran, line 79: Any reason to use `State` here at all? (rather than just having that enum be a property on the workflow and using `Void` for state?)

dhavalshreyas: Oh god! :) My brain's totally fried. 

Of course, I'll make that change.

bencochran: Hah, it happens 👍

PR #34, opened 2020-07-02
Podspecs read workflow-version from a central file

dhavalshreyas:
We now have way too many `.podspec` files. Trying to figure out a way to set version and other related properties at a central place. @amorde do let me know if there's a better way to do this!


.workflow-version
```diff
@@ -0,0 +1 @@
+$workflow_version='1.0.0.alpha1'
```
JustinDSN, line 1: I thought we were going with the name “beta”.

dhavalshreyas: We will be bumping to `beta` shortly after this release (timeline tbd).

Workflow.podspec
```diff
@@ -1,6 +1,8 @@
+load('.workflow-version')
```
amorde, line 1: using `load` can cause issues when a constant is defined and multiple podspecs are loaded into Ruby - I would use `require_relative` for this

.workflow-version
amorde, line 1: constants in Ruby tend to be `ALL_CAPS_SNAKE_CASE`

in the current implementation the `$` makes me think this is an environment variable or a shell script variable, is there a reason for the `$`?

Workflow.podspec
dhavalshreyas: I was having a tough time getting `require_relative` to work with a file name `.workflow-version`. If I name it with a `.rb` extension, it works fine.

.workflow-version
dhavalshreyas: No reason, will switch to  all caps, snake case.

AquaGeek, line 1: Should we follow [Chromium's approach](https://github.com/chromium/chromium/blob/f2247f1b3040c9831f950d1c5523ea20b3157a1a/chrome/VERSION) here and break the version into its constituent parts? That makes it _really_ easy to e.g. update from automated tooling.

Workflow.podspec
amorde: ah yes that's possible

.workflow-version
amorde: switching it will introduce issues with `load` mentioned above, so if we're sticking with `load` then we shouldn't use the all-caps version (Ruby actually treats them as constants)

dhavalshreyas: Good idea, it'll help make our automation cleaner as we keep going.

version.rb
```diff
@@ -0,0 +1,3 @@
+load('VERSION')
+
+WORKFLOW_VERSION="#{MAJOR}.#{MINOR}.#{PATCH}.#{PRE_RELEASE_IDENTIFIER}"
```
AquaGeek, line 3: Doesn't CocoaPods expect prerelease versions to look like this: `1.0.0-alpha1`?

This also needs to account for when we drop the prerelease identifier for 1.0 (though we could deal with that then).

dhavalshreyas: Yes, that last `.` should be `-`. Will fix.
Yes, I'll fix for the `PRE_RELEASE_IDENTIFIER`.

dhavalshreyas:
Builds keep failing with: 

```
Run bundle exec pod gen Development.podspec

Generating Development in `gen/Development`

[!] Unable to add a source with url `https://cdn.cocoapods.org/` named `trunk`.
[!] Automatically assigning platform `iOS` with version `11.0` on target `App-iOS` because no platform was specified. Please specify a platform for this target in your Podfile. See `https://guides.cocoapods.org/syntax/podfile.html#platform`.
You can try adding it manually in `/Users/runner/.cocoapods/repos` or via `pod repo add`.
##[error]Process completed with exit code 1.
```

@amorde: Any clue what's going on here?

dhavalshreyas:
(Works fine locally)

amorde:
Where is that happening? That’s usually a network issue when querying the server to confirm it’s a CDN-based spec repo. Trying usually fixes it

VERSION
```diff
@@ -0,0 +1,4 @@
+MAJOR=1
+MINOR=0
+PATCH=0
+PRE_RELEASE_IDENTIFIER='alpha.1'
```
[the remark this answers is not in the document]
amorde: This should be good, I believe CocoaPods mostly ignores the exact formatting of the metadata section but this is valid as far as SemVer is concerned https://semver.org

dhavalshreyas:
> Where is that happening? That’s usually a network issue when querying the server to confirm it’s a CDN-based spec repo. Trying usually fixes it

Yes, retrying worked!

PR #37, opened 2020-07-08
Fix WorkflowTesting with Multiple Child Workflows

dhavalshreyas:
`WorkflowTesting` was not accurately testing the scenario where we render multiple child Workflows with the same key.

Fixing it here.

## Checklist

- [X] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
+++ b/WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
@@ -29,6 +29,8 @@
             let file: StaticString
             let line: UInt
 
+            private var usedWorkflowKeys: Set<WorkflowKey> = []
+
             internal init(
                 state: WorkflowType.State,
                 expectedWorkflows: [AnyExpectedWorkflow],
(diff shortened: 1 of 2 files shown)


WorkflowTesting/Sources/Internal/RenderTester+TestContext.swift
```diff
@@ -117,6 +123,11 @@
                     producedOutput = output
                 }
             }
+
+            private struct WorkflowKey: Hashable {
+                let type: ObjectIdentifier
```
AquaGeek, line 128: Why `ObjectIdentifier` instead of `Any.Type`?

dhavalshreyas: `ObjectIdentifier` gets us synthesized `Equatable` and `Hashable` conformance. Not sure if there's a benefit to using `Any.Type`.

dhavalshreyas:
> This doesn't address the underlying issue with workers, but it adds tests to verify behavior. Correct?

That's right.

PR #38, opened 2020-07-08
Support keyPrefix for SignalWorkers [DNM]

dhavalshreyas:
Rough draft of what needs to be done to support propagating the `SignalWorker.key` to the `Workflow`. 

I couldn't figure out a way to accomplish this without modifying `AnyWorkflow`. I think this is too invasive for a temporary change. Unless anybody else has a better idea or strongly believe we should do this, I'll close this.


AquaGeek:
As we discussed offline, this seems like too invasive a change, and I'd be more in favor of hard-deprecating (i.e. `unavailable` instead of `deprecated`) the `asWorker` APIs instead.

PR #40, opened 2020-07-08
Mark asWorker unavailable

dhavalshreyas:
The `key` in `asWorker` is not carried over to the key used to render the `Workflows`. Since, the `Workers` are now keyed based on type + key, this may cause a run-time exception. Marking `asWorker` as `unavailable` minimize this risk.


dhavalshreyas:
> The double-firing subscription bug doesn't come back, right?

🤞🏽

It shouldn't, we're still using the same `sink` / `Action` infra to do this.

PR #41, opened 2020-07-10
Fixing Release Instructions

dhavalshreyas:
Bumping up `cocoapods-trunk`, to make use of the `--synchronous` option in `cocoapods-trunk = 1.5.0`.


dhavalshreyas:
`--synchronous` lets us push another dependent `.podspec`without waiting on the CDN to catch-up.

PR #44, opened 2020-07-15
Constrain Worker Rendering type to Void

dhavalshreyas:
@AquaGeek pointed this out during migration, `Rendering` for a `Worker` should always be `Void`.

### DIFF
--- a/.github/workflows/swift.yaml
+++ b/.github/workflows/swift.yaml
@@ -30,6 +30,7 @@ on:
       - 'WorkflowUI/**'
       - '.github/workflows/swift.yaml'
       - '.buildscript/build_swift_docs.sh'
+      - '*.swift'
 
 jobs:
   development-apps:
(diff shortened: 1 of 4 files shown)


.github/workflows/swift.yaml
```diff
@@ -30,6 +30,7 @@ on:
       - 'WorkflowUI/**'
       - '.github/workflows/swift.yaml'
       - '.buildscript/build_swift_docs.sh'
+      - '*.swift'
```
AquaGeek, line 33: What does this catch that the other globs didn't?

dhavalshreyas: Changes to `WorkflowReactiveSwift` or `WorkflowRxSwift` files were not caught.

We will need to re-evaluate this, it might be fine to run it on every PR, now that Kotlin is in it's own repo.

PR #51, opened 2020-07-24
[DNM] Ideas to make state mutations from Child Workflows easier

dhavalshreyas:
Explores introducing APIs to make State mutations easier.

## Proposal 1: 

Introduce:
```swift
extension AnyWorkflow {
    public func applyOutputToState<Parent: Workflow>(in context: RenderContext<Parent>,
                                                     stateMutation: @escaping ((inout Parent.State, Output) -> Void)) 
-> AnyWorkflow<Rendering, AnyWorkflowAction<Parent>>
}
```

that let's us do:
```swift
WelcomeWorkflow()
    .mapOutput { .demo(name: $0) }
    .applyOutputToState(in: context, keyPath: \State.self)
    .rendered(in: context)
```

## Proposal 2:

Introduce:
```swift
extension RenderContext {
    public func makeStateMutationSink() -> StateMutationSink<WorkflowType>
}
```

that let's us do:

```swift
let sink = context.makeStateMutationSink()

return FooScreen(
    title: "Leading Foo screen",
    viewTapped: {
        sink.send(\State.self, value: state + 1)
    }
)

```

### DIFF
--- a/Samples/SampleApp/Sources/RootWorkflow.swift
+++ b/Samples/SampleApp/Sources/RootWorkflow.swift
@@ -37,25 +37,6 @@ extension RootWorkflow {
     }
 }
 
-// MARK: Actions
-
-extension RootWorkflow {
-    enum Action: WorkflowAction {
-        typealias WorkflowType = RootWorkflow
-
-        case login(name: String)
-
-        func apply(toState state: inout RootWorkflow.State) -> RootWorkflow.Output? {
-            switch self {
-            case .login(name: let name):
-                state = .demo(name: name)
-            }
-
-            return nil
-        }
-    }
-}
-
 // MARK: Rendering
 
 extension RootWorkflow {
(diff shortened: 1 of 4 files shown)


bencochran:
Proposal 1 is one of the things I was thinking in the [old brain dump](https://docs.google.com/document/d/1I-HwpcmetScd5rNEvA9LCSNeEAIeYjhfYhY8fi7f3MQ/edit#heading=h.yvv8gy9l07if) (internal link, sorry external people). I think we skipped it because it wasn’t 1.0-blocking, but my `onOutput` was exactly what your `applyOutputToState` signature looks like. Not sure what your keyPath-based one is doing though, can you expand on that?

(Haven’t had a chance to look at / think about proposal 2 yet)

bencochran:
Also, just realized this is a PR not an issue, will look at the `keyPath` bit.

PR #55, opened 2020-08-20
Introduce  'makeOutputSink' convenience on 'RenderContext'

mattfaluotico:
## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/Workflow/Sources/RenderContext.swift
+++ b/Workflow/Sources/RenderContext.swift
@@ -148,4 +148,11 @@ extension RenderContext {
                 }
             }
     }
+
+    /// Generates a sink that allows sending the Workflow's output wrapped in an AnyWorkflowAction, allowing bypassing an
+    /// intermediate action.
+    public func makeOutputSink() -> Sink<WorkflowType.Output> {
+        return makeSink(of: AnyWorkflowAction.self)
+            .contraMap { AnyWorkflowAction<WorkflowType>(sendingOutput: $0) }
+    }
 }


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ _2020-07-15_
 * Fix WorkflowRxSwift podsepcs (#45)
 * Constrain Worker Rendering type to Void (#44)
 * Remove unncessary dependencies (#47)
+* Add `makeOutputSink` convenience on `RenderContext`
```
bencochran, line 12: This isn’t in alpha 4, so we shouldn’t add it here. We’ll add it to the changelog when we cut the next release.

Workflow/Sources/RenderContext.swift
```diff
@@ -148,4 +148,11 @@ extension RenderContext {
                 }
             }
     }
+
+    /// Generates a sink that allows sending the Workflow's output wrapped in an AnyWorkflowAction, allowing bypassing an
+    /// intermediate action.
+    public func makeOutputSink(of output: WorkflowType.Output) -> Sink<WorkflowType.Output> {
```
bencochran, line 154: should just be `func makeOutputSink() -> Sink<WorkflowType.Output>` since we want to be able to send the output to the resulting sink, not at creation time (as evidenced by `output` never being used here)

mattfaluotico: Oh you're totally right. 

Surprised xcode didn't alert me that the params were not used...

mattfaluotico:
Just removed the WIP. Wanted to make sure everyone liked the naming and such. It's good to go.

PR #56, opened 2020-08-22
Add `AnyWorkflow(rendering:)` to make a constant-rendering workflow

bencochran:
[Kotlin has `Workflow.rendering()`](https://github.com/square/workflow-kotlin/blob/main/workflow-core/src/main/java/com/squareup/workflow1/StatelessWorkflow.kt#L114-L120). Seems like a handy convenience for fakes and testing (and dependency injection where you support dynamic content but some consumers just want static renderings).

## Checklist

- [X] Unit Tests
- [X] ~UI Tests~
- [X] ~Snapshot Tests (iOS only)~
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/Workflow/Sources/AnyWorkflow+Rendering.swift
+++ b/Workflow/Sources/AnyWorkflow+Rendering.swift
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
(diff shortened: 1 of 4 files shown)


zach-klippenstein (approved):
We have a `Workflow.rendering(r: R): Workflow<Unit, Nothing, R>` function that does something very similar.

PR #57, opened 2020-08-25
Fixing RenderTester assert(action:)

dhavalshreyas:
In `AnyWorkflowConvertible` we were mapping all actions to `AnyWorkflowAction`. Fixing it here to map to a generic `Action` type.

## Checklist

- [x] Unit Tests


Workflow/Sources/AnyWorkflowConvertible.swift
```diff
@@ -38,7 +38,7 @@ extension AnyWorkflowConvertible {
     ///
     /// - Returns: The `Rendering` generated by the workflow.
     public func rendered<Parent>(in context: RenderContext<Parent>, key: String = "") -> Rendering where Output: WorkflowAction, Output.WorkflowType == Parent {
-        return asAnyWorkflow().render(context: context, key: key, outputMap: { AnyWorkflowAction($0) })
+        return asAnyWorkflow().render(context: context, key: key, outputMap: { $0 })
     }
 
     public func rendered<Parent, Action>(in context: RenderContext<Parent>, key: String = "", outputMap: @escaping (Output) -> Action) -> Rendering where Action: WorkflowAction, Action.WorkflowType == Parent {
```
bencochran, line 44: The line below this too, yeah? (github won’t let me comment on collapsed parts of code…)

WorkflowTesting/Tests/WorkflowRenderTesterTests.swift
```diff
@@ -98,6 +98,19 @@ final class WorkflowRenderTesterTests: XCTestCase {
             .assertNoAction()
     }
 
+    func test_childWorkflowAction() {
```
bencochran, line 101: Would this test fail before? I thought this only happened when things got `AnyWorkflow`d?

dhavalshreyas: This test did fail before the fix.

PR #60, opened 2020-09-01
Downgrading RxSwift

dhavalshreyas:


WorkflowRxSwift.podspec
```diff
@@ -19,7 +19,9 @@ Pod::Spec.new do |s|
   s.source_files = 'WorkflowRxSwift/Sources/**/*.swift'
 
   s.dependency 'Workflow', "#{s.version}"
-  s.dependency 'RxSwift', '~> 5.1.1'
+  s.dependency 'RxSwift', '= 4.4.0'
+  s.dependency 'RxCocoa', '= 4.4.0'
+  s.dependency 'RxAtomic', '= 4.4.0'
```
sebastianv1, line 24: Can we use `>= 4.4.0`?

dhavalshreyas: I don't see any breaking changes - will switch to `>=`

bencochran:
Does WorkflowRxSwift still work with 5.x correctly? If not, it’s a bummer to lock us in to 4.x for all of Workflow 1.x (as per my understanding of semver).

dhavalshreyas:
> Does WorkflowRxSwift still work with 5.x correctly? If not, it’s a bummer to lock us in to 4.x for all of Workflow 1.x (as per my understanding of semver).

There're no breaking-changes b/w 4.4.0 up to 5.x. There're a couple of deprecations - which results in warnings.

bencochran:
Thanks, mostly wanted to make sure we’re not stuck needing a 2.x over here if/when we update to RxSwift 5.x internally

WorkflowRxSwift.podspec
```diff
@@ -19,7 +19,9 @@ Pod::Spec.new do |s|
   s.source_files = 'WorkflowRxSwift/Sources/**/*.swift'
 
   s.dependency 'Workflow', "#{s.version}"
-  s.dependency 'RxSwift', '~> 5.1.1'
+  s.dependency 'RxSwift', '>= 4.4.0'
+  s.dependency 'RxCocoa', '>= 4.4.0'
+  s.dependency 'RxAtomic', '>= 4.4.0'
```
bencochran, line 24: Sorry, one more question. Do we actually use RxCocoa and RxAtomic directly?

dhavalshreyas: Nope, but it's needed for `RxSwift`. `WorkflowRxSwift` failed to compile without these.

bencochran: That’s weird. So RxSwift depends them but doesn’t declare them as dependencies? FWIW I couldn’t repro that problem locally removing these and doing `pod gen`

dhavalshreyas: `pod gen` is probably using `5.1.1`. If you could pin it to `4.4.0`, the problem repros (I think).

sebastianv1: `RxSwift` depends upon `RxAtomic`. It's surprising if it's fails to compile without `RxCocoa` if y'all aren't using its objects since it lives above `RxSwift`

dhavalshreyas: @sebastianv1: You're right, didn't need `RxCocoa`. Removing.

bencochran: Thanks! Kinda silly about the RxAtomic thing, but we can’t go back in time and fix their bug, so. Onward to RxSwift 5.x!

bencochran (approved):

WorkflowRxSwift.podspec
sebastianv1: Bug? I think it was more of a separation for atomic primitives that could be shared without consuming `RxSwift`.

bencochran: The bug being `RxSwift` doesn’t declare a dependency on a module that must be present for it to compile.

sebastianv1: Looks like it does? https://github.com/ReactiveX/RxSwift/blob/4.4.0/RxSwift.podspec#L38
Definitely a little hidden at the bottom though.

bencochran: Looks like it was this issue: https://github.com/ReactiveX/RxSwift/issues/1878

PR #79, opened 2020-10-19
Tutorial 1: Defer creation of the WelcomeView until it's needed

AquaGeek:
This came up in #76. I updated the tutorial and finished code to defer creation of the view until it's needed.

## Checklist

- [X] ~Unit Tests~
- [X] ~UI Tests~
- [X] ~Snapshot Tests (iOS only)~
- [X] I have made corresponding changes to the documentation


AquaGeek:
Should we also update the Xcode templates to have the `guard isViewLoaded else { return }` in `screenDidChange(from:previousEnvironment:)`?

PR #88, opened 2020-11-20
Introduce renderLatestOutput API

dhavalshreyas:
A common pattern in our codebase is to mutate `State` based on `Worker` output and then consume that `State` value in the `render` method. 

Here we introduce a way to convert the latest `Output` as a `Rendering` of the `Worker`. This lets us skip the intermediate value in `State` to hold this result and the `Action`s needed to mutate it.

**Before:**

```swift
struct MyWorkflow: Workflow {
    enum Action: WorkflowAction {
        typealias WorkflowType = MyWorkflow

        case update(Int)

        func apply(toState state: inout State) -> MyWorkflow.Output? {
            switch self {
            case .update(let value):
                state.fetchedData = value
            }
            return nil
        }
    }

    func render(state: State, context: RenderContext<Self>) -> Rendering {
        DataFetchWorker()
            .mapOutput { Action.update($0) }
            .rendered(in: context)

        switch state.fetchedData {
        case .loading:
            return LoadingScreen()
        case .data(let data):
            return Screen(data: data)
        }
    }
}
```

**After:**

```swift
struct MyWorkflow: Workflow {
    func render(state: State, context: RenderContext<Self>) -> Rendering {
        let data = DataFetchWorker()
            .renderLatestOutput(startingWith: .loading)
            .rendered(in: context)

        switch data {
        case .loading:
            return LoadingScreen()
        case .data(let data):
            return Screen(data: data)
        }
    }
}
```

### DIFF
--- a/WorkflowTesting/Sources/WorkflowRenderTester.swift
+++ b/WorkflowTesting/Sources/WorkflowRenderTester.swift
@@ -22,14 +22,14 @@
     import XCTest
     @testable import Workflow
 
-    extension Workflow {
+    public extension Workflow {
         /// Returns a `RenderTester` with a specified initial state.
-        public func renderTester(initialState: Self.State) -> RenderTester<Self> {
+        func renderTester(initialState: Self.State) -> RenderTester<Self> {
             return RenderTester(workflow: self, state: initialState)
         }
 
         /// Returns a `RenderTester` with an initial state provided by `self.makeInitialState()`
-        public func renderTester() -> RenderTester<Self> {
+        func renderTester() -> RenderTester<Self> {
             return renderTester(initialState: makeInitialState())
         }
     }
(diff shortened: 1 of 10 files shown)


zach-klippenstein (approved):
Nice!

AquaGeek (approved):

WorkflowTesting/Sources/WorkflowRenderTester.swift
```diff
@@ -253,18 +253,27 @@
...
             return newDictionary
         }
     }
 
+    public extension RenderTester {
+        /// Convenience to add expectation for `RenderLatestOutputWorkflow`
+        /// - Parameters:
+        ///   - producing: Value to return as `Rendering`
+        ///   - key: The key of the expected workflow (if specified).
+        func expectLatestOutputRenderingWorkflow<Value>(producing: Value, key: String = "") -> RenderTester<WorkflowType> {
```
bencochran, line 275: Thinking about this more, I’m wondering if it would be nicer for the test writer to not need to be concerned with the workflow implementation of this. Imagine a workflow like:

```swift
extension AccountLoadingWorkflow {

    func render(state: State, context: RenderContext<Self>) -> AccountLoadingScreen {
        let loadState = AccountLoadingWorker(id: id)
            .mapOutput { result -> LoadState<Account, Error> in
                switch result {
                case .success(let account):
                    return .loaded(account)
                case .failure(let error):
                    return .failed(error)
                }
            }
            .renderLatestOutput(startingWith: .loading)

        return AccountLoadingScreen(state: loadState)
    }

}
```

It’d be nice to have a test like:

```swift
AccountLoadingWorkflow(id: "b3n")
    .renderTester()
    .expect(worker: AccountLoadingWorker(id: "b3n"))
    .render { screen in
        XCTAssertEqual(screen.state, .loading)
    }
```

This lets me test that the underlying worker is actually configured as expected and leaves this feeling like an “operator” rather than another layer in the tree. I’m not sure how we’d handle non-initial values with that method, however. (I’m also not _entirely_ sure how we’d pull this off exactly, but I think we could _probably_ find a way?)

dhavalshreyas: Yes, this would've been great. This was my first thought, I dug around a bit, but didn't get too far with it.

The current `AnyWorkflow` operators do output/rendering mappings, I wasn't able to figure out a way to project outputs to renderings. If we could do something like that, we would be able to write some powerful operators (like `merge` multiple Workflows...etc).

dhavalshreyas:
Based on feedback and thinking about this more, I don't think this is ready for prime-time yet. Will move this to an internal experimental module and let it bake a little bit more.

PR #89, opened 2020-12-19
Enforce workflows are value types

kyleve:
This PR mirrors changes made in Blueprint and Listable; which add a compile time enforcement to ensure that workflows are value types. It's not foolproof; you can still of course nest a reference type in a workflow that's mutable, but this catches the top level problem.

## Checklist
- [x] Unit Tests (n/a, compile time change)
- [x] UI Tests (n/a, compile time change)
- [x] Snapshot Tests (iOS only) (n/a, compile time change)
- [x] I have made corresponding changes to the documentation (new validation type documented?)

### DIFF
--- a/Workflow/Sources/Workflow.swift
+++ b/Workflow/Sources/Workflow.swift
@@ -51,7 +51,7 @@ import ReactiveSwift
 /// }
 /// ```
 ///
-public protocol Workflow: AnyWorkflowConvertible {
+public protocol Workflow: AnyWorkflowConvertible, _Workflows_Should_Be_Value_Types {
     /// Defines the state that is managed by this workflow.
     associatedtype State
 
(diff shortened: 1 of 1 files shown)


Workflow/Sources/Workflow.swift
```diff
@@ -83,22 +83,51 @@ public protocol Workflow: AnyWorkflowConvertible {
     func render(state: State, context: RenderContext<Self>) -> Rendering
 }
 
-extension Workflow {
-    public func workflowDidChange(from previousWorkflow: Self, state: inout State) {}
+public extension Workflow {
```
kyleve, line 86: The formatter seems to have changed some things; I didn't do this one myself.

Workflow/Sources/Workflow.swift
```diff
@@ -83,22 +83,51 @@ public protocol Workflow: AnyWorkflowConvertible {
...
+/// Notes
+/// -----
+/// You should really not make your `Workflow` be a `class`. If for some reason you really really want to do this,
+/// (you should not do this unless you have a good reason, which you probably do not), then override the
+/// `workflows_should_be_value_types_by_overriding_this_method_i_acknowledge_i_am_in_hard_mode()`
+/// method in your `Workflow` to opt out of this validation – but you probably shouldn't. If you must do this,
+/// please ensure that your `Workflow` is either entirely immutable, or respects value semantics.
+///
+public protocol _Workflows_Should_Be_Value_Types {
+    func workflows_should_be_value_types_by_overriding_this_method_i_acknowledge_i_am_in_hard_mode()
```
bencochran, line 123: *extremely* nit: `by_implementing_this_method` rather than “overriding” since it’s not an `override`

Workflow/Sources/Workflow.swift
bencochran: Yeah I think at some point the default for the formatter changed or something? I dunno, but I want to bring the previous style back.

bencochran (approved):
I’ve gone back and forth a bit on if this should go on `Workflow` itself or `AnyWorkflowConvertible`. Since `AnyWorkflowConvertible` just spits out an `AnyWorkflow` that eventually leads to a `Workflow`, I think it’s okay for one of those to be a reference type (in fact, we conform `Signal` (a class) to it in `WorkflowReactiveSwift`).

Workflow/Sources/Workflow.swift
kyleve: Yeah I kinda went back and forth on this one.. You are "overriding" the default impl...

kyleve: But you're kinda not, since the method is unavailable, but then the unavailable check means it is "implemented"....

Workflow/Sources/Workflow.swift
dhavalshreyas: I guess it picked up a newer version of `SwiftFormat`. I should probably pin this to a specific version.

kyleve:
I'm gonna close this out; we added a similar validation in Blueprint, but it breaks in weird ways when extensions with where clauses are used.

PR #94, opened 2021-02-23
Handling Queued Events

dhavalshreyas:
## Queued Events
Queued Events are events sent to the `Sink` synchronously while `Screen`s are being updated. We `queue` them up and are applied after the `Screen` update operation.

## Issues
Our existing implementation of `Queued` events handling has a few issues:

* Each `Workflow` node can handle only one event, if another event is sent, we `fatalError` with a `Action sent to pipe while in the pending state.` message.
* If multiple `Workflow` nodes have events queued, only one of them will be handled, the rest are silently dropped.
* Due to a bug in `enabling` and handling `EventPipes` while handling `queued` events, the Workflow tree can get into an inconsistent state and sometimes leads to a crash.

## Solution
To solve this issue, in this PR, we are not queueing events in the `SubtreeManager`, instead we schedule the event to be delivered to the `EventPipe` asynchronously. This fixes the issues outlined above.

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -81,15 +81,6 @@ extension WorkflowNode {
         /// Enable the eventPipes for the previous rendering. The eventPipes are not valid until this has
         /// be called. If is an error to call this twice without generating a new rendering.
         func enableEvents() {
-            /// Check for queued events. If there are any, apply the first and yield to the next render loop.
-            let queuedEvents = eventPipes.compactMap { pipe in
-                pipe.pendingOutput()
-            }
-            if !queuedEvents.isEmpty {
-                handle(output: queuedEvents[0])
-                return
-            }
-
             /// Enable all action pipes.
             for eventPipe in eventPipes {
                 eventPipe.enable { [weak self] output in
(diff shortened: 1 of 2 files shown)


Workflow/Sources/SubtreeManager.swift
```diff
@@ -315,10 +313,7 @@ extension WorkflowNode.SubtreeManager {
                 fatalError("[\(WorkflowType.self)] Sink sent an action inside `render`. Sinks are not valid until `render` has completed.")
 
             case .pending:
-                validationState = .queued(event)
-
-            case .queued:
-                fatalError("[\(WorkflowType.self)] Action sent to pipe while already in the `queueing` state.")
+                fatalError("[\(WorkflowType.self)] Action sent to pipe while in the `pending` state.")
```
rjrjr, line 316: I thought this PR was to eliminate this error?

dhavalshreyas: Yeah, this should not be possible, however the `.pending` state will continue to exist. `fatal` ensures we don't hit this spot.

rjrjr:
Posthoc LGTM

rjrjr (approved):

AquaGeek (approved):
Have we run this against the tests in the monorepo?

dhavalshreyas:
> Have we run this against the tests in the monorepo?

Yes, all green.

PR #96, opened 2021-03-31
Bump RxSwift to 5.1

dhavalshreyas:
Internally, Square has switched to `RxSwift 5.1.1`. This change reflects that update.


asthatrivedi (approved):

AquaGeek:
Can you provide more context here? Why 5.1.1?

PR #100, opened 2021-09-06
Updates to Workflow Templates

monishsyed:
* Added ability to optionally create a worker or avoid the worker noise completely if your workflow isn't doing any sync work.
* Worker template now supports both RxSwift as well as ReactiveSwift
* Reusing default workflow template using symlinks to avoid duplication

Screen Shot 2021-09-07 at 1 26 54 AM


Tooling/Templates/Workflow (Verbose).xctemplate/generateWorkerRxSwift/___FILEBASENAME___Workflow.swift
```diff
@@ -0,0 +1 @@
+/Users/monishsyed/Development/Personal/workflow-swift/Tooling/Templates/Workflow (Verbose).xctemplate/Default/___FILEBASENAME___Workflow.swift
```
dhavalshreyas, line 1: Do we need this file? The pathname should be relative.

Tooling/Templates/Workflow (Verbose).xctemplate/TemplateInfo.plist
```diff
@@ -18,12 +18,52 @@
...
+			<key>Required</key>
+			<true/>
+			<key>Name</key>
+			<string>Uses:</string>
+			<key>Description</key>
+			<string>Whether to use RxSwift or ReactiveSwift</string>
+			<key>Type</key>
+			<string>popup</string>
+			<key>Default</key>
+			<string>RxSwift</string>
```
AquaGeek, line 53: I think the default should be `ReactiveSwift`. @dhavalshreyas thoughts?

Tooling/Templates/Workflow (Verbose).xctemplate/generateWorkerRxSwift/___FILEBASENAME___Workflow.swift
monishsyed: This is the symlink for the original file that resides under `Tooling/Templates/Workflow (Verbose).xctemplate/Default/`. Even though I had created the symlink with the relative path, seems like symlinks cannot resolve relative paths and so symlinks have to have absolute path ALWAYS.

> Why was symlink created in the first place?

It was created to overcome a limitation that we have with xcode files templates i.e files templates cannot be inherited from other templates (project templates can have ancestors though but not file templates) and therefore when creating a template that has RxSwift/ReactiveSwift worker enabled, both `___FILEBASENAME___Workflow.swift` and `__FILEBASENAME___Worker.swift` has to be provided inside `generateWorkerRxSwift` and `generateWorkerReactiveSwift`. This leaves us in a situation where `___FILEBASENAME___Workflow.swift` will have 3 identical copies and any changes to `___FILEBASENAME___Workflow.swift` would require updates to all 3. `Symlinks` was one solution since it allows you to create aliases for these files. But as you can see, symlinks don't work with relative paths unfortunately :(

> What are other ideas to avoid duplication of `___FILEBASENAME___Workflow.swift`?

1. Maybe just have one copy `___FILEBASENAME___Workflow.swift ` in `Default` and run copy commands as part of `install-xcode-templates.sh` script that copies it in `generateWorkerRxSwift` and `generateWorkerReactiveSwift`. Downside is without running the script, the template is pretty much broken. Ideally script should only be there to help you install templates.
2. Generate the full templates through scripts. May need a bit of work there.
3. Keep 3 copies of  `___FILEBASENAME___Workflow.swift` if we don’t anticipate changes to the templates often. I think it should ok for 1.0.0. 


What do you all think?

Tooling/Templates/Workflow (Verbose).xctemplate/TemplateInfo.plist
dhavalshreyas: Yes, `ReactiveSwift` has been our default.

Tooling/Templates/Workflow (Verbose).xctemplate/generateWorkerRxSwift/___FILEBASENAME___Workflow.swift
monishsyed: Went ahead and added back copies of `___FILEBASENAME___Workflow.swift` for now

Tooling/Templates/Workflow (Verbose).xctemplate/TemplateInfo.plist
monishsyed: Changed default to `ReactiveSwift `

dhavalshreyas (approved):
This looks great, thanks for doing this @monishsyed!

At some point, we should come back and add tests here to make sure we're not breaking something as we iterate.

dhavalshreyas (approved):

PR #107, opened 2021-12-04
Make build and update method on VC description public.

kyleve:
In many places in Market, it would be useful to directly manage our view controller content, without needing to wrap view controllers into a `DescribedViewController`, eg for example, in `MarketNavigationController`, where the extra `DescribedViewController` wrapper provides no benefit, since screens of different types are considered different screens by the nav controller.  This also provides performance benefits; not creating an extra wrapper view controller for every one shown in a nav stack.

By making build, update, canUpdate, public, we can remove the extra `DescribedViewController` layer; and manage these VCs directly

We'd also like to eventually use these newly public'd methods to remove lots of extra VC layering from modals and other areas in Market.

There's additional cases where making these methods public is useful: Now, a `Screen` can describe ` UIActivityViewController` (share sheet) to present, which cannot be wrapped in a parent during presentation for the presentation to work correctly, for example.

## Checklist

- [x] Unit Tests (Not relevant)
- [x] UI Tests (Not relevant)
- [x] Snapshot Tests (iOS only) (Not relevant)
- [x] I have made corresponding changes to the documentation (Updated code API docs)


WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -18,8 +18,12 @@
...
     import UIKit
 
-    /// A ViewControllerDescription acts as a recipe for building and updating a
-    /// specific UIViewController.
+    /// A ViewControllerDescription acts as a recipe for building and updating a specific UIViewController.
+    ///
+    /// You usually use a `ViewControllerDescription` in conjunction with a `DescribedViewController`,
+    /// which will internally create and manage the described view controller for its current view controller description.
+    /// However, you can also directly invoke the public methods such as `buildViewController()`,
+    /// `update(viewController:)`, if you are manually managing your own view controller hierarchy.
```
n8chur, line 26: I'd probably be good to first describe what I'd imagine is the most frequently use of this interface: providing a description of the backing view in a `Screen`. I typically see `Screen`s, not `ViewControllerDescription`s, passed to `DescribedViewController`.

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -70,7 +76,16 @@
         ///         `canUpdate(viewController:)` will result in an exception.
         ///
         /// - Parameter viewController: The view controller instance to update
```
n8chur, line 78: We may want to also add a note about the precondition being hit if the type is wrong here.

n8chur (approved):

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
kyleve:
```
Note: Passing a view controller that does not return `true` from
        ///         `canUpdate(viewController:)` will result in an exception.
```

Is in the note section, I can re-word that

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
kyleve: Sounds good! I can drop in some code samples

dhavalshreyas:
In general, the change makes sense to me. However, correctly updating the `UIViewController` to reflect the changes can prone to errors and will need to be carefully managed. Having this be private will hide that complexity, at the cost of the additional VC.

Would love to hear @bencochran 's thoughts on this.

kyleve:
Yep; definitely requires some carefulness – I wouldn't recommend most folks reach for this – but removing the many extra layers of wrapper VCs is a fairly large win in terms of debuggability of the view controller hierarchy, and I can imagine the lack of all those VCs will add up in terms of mem usage as well (albeit I haven't verified the exact allocation cost of a VC recently; it used to be pretty substantial).

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
kyleve: Took a shot at updating the docs of use cases before adding code instead.

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
kyleve: Done

kyleve:
Bumping this!

kyleve:
Boink

dhavalshreyas (approved):
"more power, more responsibility". Agreed with that line, let's try it out!

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -46,31 +59,46 @@
...
         ///
-        /// - Note: Passing a view controller that does not return `true` from
-        ///         `canUpdate(viewController:)` will result in an exception.
+        /// - Parameters:
+        ///   - viewController: The view controller to update.
         ///
-        /// - Parameter viewController: The view controller instance to update
-        internal func update(viewController: UIViewController) {
+        /// ### Note
+        /// You must pass a view controller that is exactly the same type as the type passed to `init`'s `type`
```
bencochran, line 89: I'd prefer to keep this a little more implementation-vague with something like 

> You must pass a view controller previously created by a compatible ViewControllerDescription and that passes `canUpdate(viewController:)`. Failure to do so will result in a fatal `precondition`.

rather than talking about the type itself (slightly leaving the door open for further refinements internally here if they're ever needed).

bencochran (approved):
I'm generally hesitant to give this potential foot gun, but I do agree that there's a need from some consumers to have a lighter-weight way to contain a screen.

I'd also love some quantitive analysis of the actual overhead of view controller wrappings (vs the very understandable "that can't be good" gut reaction to seeing a ridiculously deep tree).

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
kyleve: Done!

PR #108, opened 2021-12-08
Fix bugs related to view controller containment

n8chur:
# Overview

- Adds additional view controller property forwarding to contained view controllers in `ContainerViewController` and `DescribedViewController`
- Fixes a missing preferred content size update and avoids updating the preferred content size in a few cases where it did not previously
- Adds contained child view controllers as children in the initializer for these containers. This allows consumers to inspect properties of the view controller before the view has loaded and get appropriate values—previously this would return a default value until the view was loaded even though the information was available.

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowUI/Sources/Container/ContainerViewController.swift
+++ b/WorkflowUI/Sources/Container/ContainerViewController.swift
@@ -78,6 +78,8 @@
             addChild(rootViewController)
             view.addSubview(rootViewController.view)
             rootViewController.didMove(toParent: self)
+
+            updatePreferredContentSizeIfNeeded()
         }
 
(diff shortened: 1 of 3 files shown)


kyleve (approved):

n8chur:
- [ ] Move view controller containment (addChild etc.) to initializers

n8chur:
This containment / `preferredContentSize` pattern has been used in Market and has proven to work well for us so far!

dhavalshreyas (approved):

amorde (approved):

PR #112, opened 2021-12-17
Add WrapperScreen

n8chur:
`WrapperScreen` embeds a wrapped `Screen`'s backing view controller as a child view controller with a matching frame.

This `Screen` can be useful when you'd like to add view controller functionality to an existing `Screen` when you don't have access to the backing view controller. For example, you might use a `WrapperScreen` to present a share  sheet over an existing `Screen`.

`WrapperScreen` is intended for use with a `WrapperScreenViewController` subclass. `WrapperScreenViewController` handles embedding the wrapped `Screen`, adding it's backing view controller as a child, and updating its frame to match the the backing view, forwards child view controller methods, and updates the wrapped screen when appropriate.

 Provide a view controller description of your `WrapperScreenViewController` subclass in your conformance to `WrapperScreen`:

```swift
 struct MyWrapper {
     var wrapped: Content
 }

 extension ShareSheetScreen: WrapperScreen, Screen where Content: Screen {
     public func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
         MyWrapperScreenViewController<Content>.description(for: self, environment: environment)
     }
 }
 ```

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### CODE REVIEW

File: Wrapper/WrapperScreen.swift [hunk 1]
```diff
@@ -0,0 +1,134 @@
...
+     extension ShareSheetScreen: WrapperScreen, Screen where Content: Screen {
+         public func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
+             MyWrapperScreenViewController<Content>.description(for: self, environment: environment)
+         }
+     }
+     ````
+
+     See `WrapperScreenViewController` for more information on how write your view controller subclass.
+     */
+    public protocol WrapperScreen: Screen {
```
kyleve, line 33: Naming bikeshed: `ContainerScreen`?

Wrapper/WrapperScreen.swift
```diff
@@ -0,0 +1,134 @@
...
+        /// You must call `super.screenDidChange(from: previousScreen, previousEnvironment: previousEnvironment)` in
+        /// your implementation.
+        override open func screenDidChange(from previousScreen: ScreenType, previousEnvironment: ViewEnvironment) {
+            super.screenDidChange(from: previousScreen, previousEnvironment: previousEnvironment)
+
+            wrappedViewController.update(screen: screen.wrapped, environment: environment)
+        }
+
+        override open func loadView() {
+            view = UIView(frame: UIScreen.main.bounds)
```
kyleve, line 68: We should also apply this, as @kylebshr found:

```
        // Match the UIKit default for a view controllers view. Not doing so can cause unexpected layout behavior.
        view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```

kyleve (approved):

Wrapper/WrapperScreen.swift
kylebshr: I do think this is worth doing, in most cases it doesn't affect Square code since we don't really _use_ autoresizing masks, but for an open source framework it seems good to do (and in case that ever changes at Square)

Wrapper/WrapperScreen.swift
kylebshr: Dang I like this name a lot better but I'm worried that `ContainerViewController` overloads this term already and people will assume this is a screen backed by `ContainerViewController` or that they're somehow related.

Maybe it's fine, since each of the types are documented and have clearly different purposes?

kylebshr (approved):

Wrapper/WrapperScreen.swift
n8chur: Yeah I avoided `ContainerScreen` due to `ContainerViewController`—I also think that sharing the same prefix would make it confusing.

I kinda wish `ContainerViewController` was `WorkflowViewController`, but that rename seems like a bit too much of a breaking change.

Open to alternatives!

Wrapper/WrapperScreen.swift
n8chur: - [ ] Set `autoresizingMask` on views

Wrapper/WrapperScreen.swift
kyleve: I wonder how painful it would be to rename `ContainerViewController`. I have always disliked the name and agree that `WorkflowViewController` or `WorkflowHostViewController` would be better.

Looking at our repo:

```
git grep "ContainerViewController(workflow:" | wc -l
     180
```

All these seem like they could be simple finds and replaces. We could do it in two phases (rename, deprecated typealias, move over, remove typealias) fairly easily.

n8chur: I'm open to it! Curious how others feel.

Also note this would lead to a 2.0.0 when we do decide to remove the typealiases (which seems totally fine to me!).

kylebshr: I'm for this 👍

kylebshr: I wonder if that would mean changing https://github.com/square/workflow-kotlin to match anywhere?

dhavalshreyas: Well, we've gotten really good at doing such migrations, using `SwiftSyntax`.

Not sure how I feel about the `Workflow` prefix, however `HostViewController` sounds good.

dhavalshreyas:
Would love to sync up with @n8chur to understand the motivation for this. Also, not sure where this would go, looks like it's currently not hosted in any pod/package.

n8chur:
> not sure where this would go, looks like it's currently not hosted in any pod/package.

Oops, looks like I got the folder path wrong. It's meant to live in `WorkflowUI/Sources/Wrapper` and be distributed with that package.

- [ ] Move source to appropriate folder

> Would love to sync up with @n8chur to understand the motivation for this

Happy to schedule time to chat if you'd prefer, but in short: we've found it useful to create wrapper screens that extend functionality to a backing view controller and doing so requires a lot of very easy to mess up boilerplate code related to view controller containment and managing the embedded screen.

An example of this would be Register's [ShareSheetScreen](https://github.com/squareup/ios-register/tree/master/Frameworks/ShareSheetScreen) which allows adding native UIKit share sheet functionality to an existing screen without needing access to the embedded `Screen`'s backing view controller. You can find an example of how this convenience simplifies usage of this pattern in [this example PR](https://github.com/squareup/ios-register/pull/64272/files) which updates `ShareSheetScreen` to utilize the `WrapperScreen` / `WrapperScreenViewController`.

View controller containment in general is fragile and very easy to mess up—we find bugs related to it extremely often throughout Register. Rather than scattering this complexity throughout our codebase it seems it could be appropriate to add this convenience directly to `WorkflowUI`!

n8chur:
Spoke with @dhavalshreyas and new plan is to try this out in `WorkflowExperimental` for now. I don't think we have a direct usecase in Market just yet so it should be fine that it lives in Register for now!

If we find uses for it in Market we can move it to a separate framework next to `Workflow` (`WorkflowUtils`?) or add it directly to the Market

PR #115, opened 2022-01-08
Always remove child view controller from DescribedViewController

kylebshr:
Fixes an issue where the child of a `DescribedViewController` wasn't removed even when the `currentViewController` was replaced. This issue was introduced in https://github.com/square/workflow-swift/pull/108 - we started setting up the parent-child relationship in the init, but only remove the child if our view is loaded. In the case where an update happens before the view is loaded, that initial child is never removed.

The left-behind relationship and view was causing issues where (correct) assumptions about the view and view controller hierarchy were invalid, causing this crash: https://square.slack.com/archives/CTJ4UNLHF/p1641510705024100

Thanks @tyten and @kyleve for helping debug!

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowUI/Tests/DescribedViewControllerTests.swift
+++ b/WorkflowUI/Tests/DescribedViewControllerTests.swift
@@ -108,10 +108,11 @@
 
             // Then
             XCTAssertNotEqual(initialChildViewController, describedViewController.currentViewController)
+            XCTAssertNil(initialChildViewController.parent)
             XCTAssertEqual((describedViewController.currentViewController as? MessageViewController)?.message, "Test")
             XCTAssertFalse(describedViewController.isViewLoaded)
             XCTAssertFalse(describedViewController.currentViewController.isViewLoaded)
-            XCTAssertNil(describedViewController.currentViewController.parent)
+            XCTAssertEqual(describedViewController.currentViewController.parent, describedViewController)
         }
 
         func test_update_toIncompatibleDescription_afterViewLoads() {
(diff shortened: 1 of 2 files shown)


kyleve (approved):

WorkflowUI/Tests/DescribedViewControllerTests.swift
```diff
@@ -108,10 +108,11 @@
 
             // Then
             XCTAssertNotEqual(initialChildViewController, describedViewController.currentViewController)
+            XCTAssertNil(initialChildViewController.parent)
             XCTAssertEqual((describedViewController.currentViewController as? MessageViewController)?.message, "Test")
             XCTAssertFalse(describedViewController.isViewLoaded)
             XCTAssertFalse(describedViewController.currentViewController.isViewLoaded)
-            XCTAssertNil(describedViewController.currentViewController.parent)
+            XCTAssertEqual(describedViewController.currentViewController.parent, describedViewController)
```
kylebshr, line 115: We actually do want this child-parent relationship set up, which is also fixed

WorkflowUI/Tests/DescribedViewControllerTests.swift
```diff
@@ -108,10 +108,11 @@
 
             // Then
             XCTAssertNotEqual(initialChildViewController, describedViewController.currentViewController)
+            XCTAssertNil(initialChildViewController.parent)
```
kylebshr, line 111: This test validates the fix. It fails on `main`.

tyten (approved):

dhavalshreyas (approved):
Thank you!

n8chur (commented):
Thanks for fixing!

PR #122, opened 2022-02-22
Add listeners and remove ReactiveSwift exposed types

mjohnson12:
This PR represents one way the ReactiveSwift properties on `WorkflowHost` and `ContainerViewController` could be removed and replaced with the ability to add listeners for rendering and output.

One of the requirements was to expose properties on `WorkflowHost` in the specific FRP module for rendering and output.
I.E. WorkflowReactiveSwift would have an extension on `WorkflowHost` to expose a ReactiveSwift Signal property for rendering and output.
A listener is a class that provides the ability for consumers to be notified when the workflow renders or when output occurs.  A closure listener is provided in the base Workflow module as well as function overloads on `WorkflowHost` to be able to add a closure listener for rendering or output.
Another requirement was to be able to support multiple listeners.  WorkflowHost supports a collection of listeners for rendering and output.
Each FRP specific module provides a listener subclass that implements properties for rendering and output.

A class was chosen for the listener because the type needs to be generic to Rendering or Output of the Workflow.  Using a protocol would have forced Workflow to be generic over the listener because of using an associated type in the protocol.  The class hierarchy makes it easy to add more listeners since you just subclass listener to add a new listener type.

The current implementation of `WorkflowHost` exposes ReactiveSwift types such as Property and Signal.  The behavior of these types is specific to ReactiveSwift.  Removing these types changes the behavior of `WorkflowHost` and `ContainerViewController`.
We might be better off waiting until we can support iOS 13+ and use Combine in Workflow since it will provide similar functionality to that which is exposed now.

## Checklist

- [X] Unit Tests
- [ ] UI Tests


Workflow/Sources/DispatchQueue+Workflow.swift
```diff
@@ -15,16 +15,7 @@
  */
 
 import Foundation
-import ReactiveSwift
 
 extension DispatchQueue {
     static let workflowExecution: DispatchQueue = .main
```
JustinDSN, line 20: I assume this static property is still referenced somewhere in the Workflow framework? It's marked as internal.

Workflow/Sources/WorkflowHost.swift
```diff
@@ -80,11 +78,69 @@ public final class WorkflowHost<WorkflowType: Workflow> {
         handle(output: output)
     }
 
+    /// Add a listener closure to be called back when a render pass happens.
+    public func addRenderingListener(_ closure: @escaping (WorkflowType.Rendering) -> Void) -> UUID {
+        dispatchPrecondition(condition: .onQueue(DispatchQueue.main))
```
JustinDSN, line 83: Just want to verify we do not want to be on `DispatchQueue.workflowExecution`?

mjohnson12: Dhaval said that workflow host should always be called on the main thread so that's why I used `main` instead of `workflowExecution` since that could be changed in the future to something else.  Also `workflowExecution` is used internally in Workflow to process node updates and these public functions are called from external sources.
Since currently these equate to the same thing I'm open to change it but I was assuming `workflowExecution` was more of an internal thing to workflow.

WorkflowReactiveSwift/Sources/WorkflowHost+Extension.swift
```diff
@@ -0,0 +1,57 @@
...
+extension WorkflowHost {
+    public var renderingSignal: Signal<WorkflowType.Rendering, Never> {
+        if let signalListener = getRenderingListener(id: WorkflowHostSignalListenerIdentifier.id) as? SignalListener {
+            return signalListener.signal
+        } else {
+            let listener = SignalListener<WorkflowType.Rendering>(id: WorkflowHostSignalListenerIdentifier.id)
+            addRenderingListener(listener: listener)
+            return listener.signal
+        }
+    }
```
dhavalshreyas, line 46: This may be better as a `SignalProducer`.  We should also handle the signal termination events.

```suggestion
    public var renderingSignalProducer: SignalProducer<WorkflowType.Rendering, Never> {
        let signalProducer: SignalProducer<WorkflowType.Rendering, Never> = SignalProducer { (observer, lifetime) in
          let listenerUUID = addRenderingListener { rendering in
               observer.send(rendering)
          }
          
          lifetime.observeEnded {
               removeListener(listenerUUID)
          }
    }
```

Workflow/Sources/WorkflowHost.swift
```diff
@@ -94,9 +150,31 @@ public final class WorkflowHost<WorkflowType: Workflow> {
 
         rootNode.enableEvents()
     }
+}
+
+// MARK: - Separate listener types
+
+open class Listener<OutputType> {
```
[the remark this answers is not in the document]
mjohnson12: We can't and be able to subclass them in the FRP specific modules.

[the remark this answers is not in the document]
dhavalshreyas: Would we need to?

Workflow/Sources/WorkflowHost.swift
```diff
@@ -80,11 +78,69 @@ public final class WorkflowHost<WorkflowType: Workflow> {
         handle(output: output)
     }
 
+    /// Add a listener closure to be called back when a render pass happens.
+    public func addRenderingListener(_ closure: @escaping (WorkflowType.Rendering) -> Void) -> UUID {
```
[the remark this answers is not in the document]
mjohnson12: It actually broke tests behavior to do that.  Before people could subscribe to a signal created from the rendering property which would only fire when rendering changed.  If we call it here it will fire immediately.
They still have the option of taking the current rendering from the property on WorkflowHost and call the listener with it themselves to trigger it.

WorkflowReactiveSwift/Sources/WorkflowHost+Extension.swift
mjohnson12: This creates a new SignalProducer for every call of the property instead of returning the same one every time.

Workflow/Sources/WorkflowHost.swift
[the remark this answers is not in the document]
mjohnson12: Yes.  The Listener subclasses implement the FRP specific functionality.  For example the PublisherListener for Workflow combine.

WorkflowReactiveSwift/Sources/WorkflowHost+Extension.swift
dhavalshreyas: Good point, would that be a problem?

mjohnson12: Since we can service more than one subscription with one instance it's more efficient to just create one listener.  The only problem to me is it's not completely obvious to the consumer because it's a property that they are creating a new instance every time they use the property.

PR #123, opened 2022-02-28
Workflow concurrency support for workers

mjohnson12:
Creating a PR for Soo's work on adding concurrency support for async workers.

This adds a WorkflowConcurrency module that defines a `Worker` protocol that can run an `async` function and allows `Task`'s to be run as a workflow.

## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowConcurrency/Sources/TaskWorkflow.swift
+++ b/WorkflowConcurrency/Sources/TaskWorkflow.swift
@@ -21,7 +21,7 @@ import Workflow
 @available(iOS 13.0, macOS 10.15, *)
 extension Task: AnyWorkflowConvertible where Failure == Never {
     public func asAnyWorkflow() -> AnyWorkflow<Void, Success> {
-        TaskWorkflow(taskProvider: { self }).asAnyWorkflow()
+        TaskWorkflow(task: self).asAnyWorkflow()
     }
 }
 
@@ -31,18 +31,17 @@ struct TaskWorkflow<Value>: Workflow {
     public typealias State = Void
     public typealias Rendering = Void
 
-    var taskProvider: () -> Task<Value, Never>
+    private let task: Task<Value, Never>
 
-    public init(taskProvider: @escaping () -> Task<Value, Never>) {
-        self.taskProvider = taskProvider
+    public init(task: Task<Value, Never>) {
+        self.task = task
     }
 
     public func render(state: State, context: RenderContext<TaskWorkflow>) -> Rendering {
         let sink = context.makeSink(of: AnyWorkflowAction.self)
-        context.runSideEffect(key: "") { [taskProvider] lifetime in
-            let providedTask = taskProvider()
-            let task = Task {
-                let output = await providedTask.value
+        context.runSideEffect(key: "") { [task] lifetime in
+            let sideEffectTask = Task {
+                let output = await task.value
                 if Task.isCancelled { return }
                 let action = AnyWorkflowAction<TaskWorkflow>(sendingOutput: output)
                 DispatchQueue.main.async {
@@ -51,8 +50,8 @@ struct TaskWorkflow<Value>: Workflow {
             }
 
             lifetime.onEnded {
+                sideEffectTask.cancel()
                 task.cancel()
-                providedTask.cancel()
             }
         }
     }
--- a/WorkflowConcurrency/Sources/TaskWorkflow.swift
+++ b/WorkflowConcurrency/Sources/TaskWorkflow.swift
@@ -18,14 +18,14 @@ import Combine
 import Foundation
 import Workflow
 
-@available(iOS 15.2, macOS 11.3, *)
+@available(iOS 13.0, macOS 10.15, *)
 extension Task: AnyWorkflowConvertible where Failure == Never {
     public func asAnyWorkflow() -> AnyWorkflow<Void, Success> {
         TaskWorkflow(taskProvider: { self }).asAnyWorkflow()
     }
 }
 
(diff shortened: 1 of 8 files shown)


WorkflowConcurrency/Tests/TaskTests.swift
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Combine
+import Foundation
+import Workflow
+// import WorkflowConcurrencyTesting
```
soorinpark, line 20: Should this be removed?

soorinpark:
I think there are a few extra test cases that other Worker frameworks cover-- might be good to add those here as well!

WorkflowConcurrency/Tests/TaskTests.swift
mjohnson12: Yes.  Removing.

mjohnson12:
I'm going to wait until testing gets added to the package via this PR https://github.com/square/workflow-swift/pull/124 to put in the additional tests so that I can use WorkflowTesting in the concurrency tests with SPM.

soorinpark (approved):

WorkflowConcurrency/Tests/WorkerTests.swift
```diff
@@ -0,0 +1,204 @@
...
+                let endExpectation: XCTestExpectation
+
+                func run() async -> Void {
+                    startExpectation.fulfill()
+                    for _ in 1 ... 4 {
+                        if Task.isCancelled {
+                            endExpectation.fulfill()
+                            return
+                        }
+                        try? await Task.sleep(nanoseconds: 500000000)
```
dhavalshreyas, line 116: Can we figure out a way to test this without an explicit sleep?

WorkflowConcurrency/Tests/WorkerTests.swift
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+@available(iOS 13.0, macOS 10.15, *)
+private struct TaskTestWorker: Worker {
+    typealias Output = Int
+
+    func run() async -> Int {
+        do {
+            try await Task.sleep(nanoseconds: 3000000000)
```
dhavalshreyas, line 197: Same here.

dhavalshreyas (commented):
This is so freakin' cool!

WorkflowConcurrency/Tests/WorkerTests.swift
mjohnson12: We don't control what thread/scheduler the async Task runs on so I don't know of a way of moving the scheduler forward in time for this.  The test needs the async function to be busy longer than the timeout so that we can verify cancellation actually works.  We have to cancel out of work that would normally go longer than the timeout.
The way the test is currently coded it should only wait for the first sleep so I can lower that to something very quick and try that so that the test runs faster.

WorkflowConcurrency/Tests/WorkerTests.swift
```diff
@@ -0,0 +1,204 @@
...
+        )
+
+        let expectation = XCTestExpectation()
+        let disposable = host.rendering.signal.observeValues { rendering in
+            expectation.fulfill()
+        }
+
+        XCTAssertEqual(0, host.rendering.value)
+
+        wait(for: [expectation], timeout: 5.0)
```
dhavalshreyas, line 36: Can we bump these down to 1?

WorkflowConcurrency/Sources/TaskWorkflow.swift
```diff
@@ -0,0 +1,59 @@
...
+    public init(taskProvider: @escaping () -> Task<Value, Never>) {
+        self.taskProvider = taskProvider
+    }
+
+    public func render(state: State, context: RenderContext<TaskWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: "") { [taskProvider] lifetime in
+            let providedTask = taskProvider()
+            let task = Task {
+                let output = await providedTask.value
```
dhavalshreyas, line 45: Can we handle `AsyncSequence` here?

WorkflowConcurrency/Tests/WorkerTests.swift
mjohnson12: Sure

JustinDSN (approved):
LGTM

WorkflowConcurrency/Sources/TaskWorkflow.swift
```diff
@@ -0,0 +1,59 @@
...
+
+    var taskProvider: () -> Task<Value, Never>
+
+    public init(taskProvider: @escaping () -> Task<Value, Never>) {
+        self.taskProvider = taskProvider
+    }
+
+    public func render(state: State, context: RenderContext<TaskWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: "") { [taskProvider] lifetime in
```
JustinDSN, line 42: Should this be “”?

mjohnson12: The key is just used to make sure you don't run the same side affect twice on the render pass.  Empty string should work here.

PR #124, opened 2022-03-01
Adds testing targets and libraries to SPM

pbelonsquare:


Package.swift
```diff
@@ -71,6 +83,21 @@ let package = Package(
             dependencies: ["Workflow"],
             path: "WorkflowCombine/Sources"
         ),
+        .testTarget(
```
pbelonsquare, line 86: Should these be regular targets?

Package.swift
```diff
@@ -30,6 +30,18 @@ let package = Package(
             name: "WorkflowCombine",
```
kyleve, line 30: Does this also need to be added to CI, or will it run already?

mjohnson12 (commented):
As mentioned in Slack the Workflow package doesn't build for testing for me.

Package.swift
pbelonsquare: Good question. I'm not familiar with the Github CI infra perhaps @mjohnson12 you have a better idea?

Package.swift
```diff
@@ -30,6 +30,18 @@ let package = Package(
...
+            name: "WorkflowReactiveSwiftTesting",
+            targets: ["WorkflowReactiveSwiftTesting"]
+        ),
+        .library(
+            name: "WorkflowCombineTesting",
+            targets: ["WorkflowCombineTesting"]
+        ),
     ],
     dependencies: [
         .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", from: "6.3.0"),
```
lickel, line 47: `.package(url: "https://github.com/nicklockwood/SwiftFormat", .exact("0.44.14")),`

Is there a reason we are pinning the SwiftFormat version?
Should generally use `from` and then depend on a Package.resolved file

Package.swift
mjohnson12: I believe the CI build using the podspec for running the tests which doesn't including anything that requires iOS 13+ since cocoapods would require workflow to require iOS 13+ so those are commented out in Development.podspec.

There is one CI test that runs with package manager but I don't know which tests it runs.

mjohnson12: Looking at the output of the swift package manager CI step and it does not run the WorkflowCombine tests.

Package.swift
mjohnson12: @dhavalshreyas Do you know what this dependency is pinned at an exact version?

Package.swift
mjohnson12: I've pushed up an update that adds WorkflowCombine to the tests that run for the SPM CI build step.

mjohnson12 (approved):

PR #151, opened 2022-08-17
[performance]: disable logging by default & add logging configuration

jamieQ:
## Background

- during some recent investigations into rendering performance, we found that leaving the runtime os logging enabled (the current implicit behavior) had a noticeable adverse effect.

## Changes

- logging is now opt-in, and disabled by default
- added additional configuration options to control which categories of runtime data are logged
- added synthetic performance tests to gut check execution time changes under various logging configurations

example XCTest run after disabling logging:
Screen Shot 2022-08-17 at 14 30 31

## Checklist

- [x] Unit Tests
- [x] TODO: apply changes to assorted WorkerLogger implementations

### DIFF
--- a/Workflow/Sources/WorkflowLogger.swift
+++ b/Workflow/Sources/WorkflowLogger.swift
@@ -14,10 +14,64 @@
  * limitations under the License.
  */
 
-import os.signpost
+import OSLog
 
 private extension OSLog {
+    /// Logging will use this log handle when enabled
     static let workflow = OSLog(subsystem: "com.squareup.Workflow", category: "Workflow")
+
+    /// The active log handle to use when logging. Defaults to the shared `.disabled` handle.
+    static var active: OSLog = .disabled
+}
+
+// MARK: -
+
+/// Namespace for specifying logging configuration data.
+public enum WorkflowLogging {
+    public struct Config {
+        /// Configuration
+        public enum RenderLoggingMode {
+            /// No data will be recorded for WorkflowNode render timings.
+            case none
+
+            /// Render timings will only be recorded for root nodes in a Workflow tree.
+            case rootsOnly
+
+            /// Render timings will be recorded for all nodes in a Workflow tree.
+            /// N.B. performance may be noticeably impacted when using this option.
+            case allNodes
+        }
+
+        public var renderLoggingMode: RenderLoggingMode = .allNodes
+
+        /// When `true`, the interval spanning a WorkflowNode's lifetime will be recorded.
+        public var logLifetimes = true
+
+        /// When `true`, action events will be recorded.
+        public var logActions = true
+    }
+
+    /// Global setting to enable or disable logging.
+    /// Note, this is independent of the specified `config` value, and simply governs whether
+    /// the runtime should emit any logs.
+    public static var enabled: Bool {
+        get { OSLog.active === OSLog.workflow }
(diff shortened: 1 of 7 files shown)


Workflow/Sources/WorkflowLogger.swift
```diff
@@ -14,10 +14,64 @@
...
 
-import os.signpost
+import OSLog
 
 private extension OSLog {
+    /// Logging will use this log handle when enabled
     static let workflow = OSLog(subsystem: "com.squareup.Workflow", category: "Workflow")
+
+    /// The active log handle to use when logging. Defaults to the shared `.disabled` handle.
+    static var active: OSLog = .disabled
```
mjohnson12, line 24: Just for my own clarification.  This defaulting to `.disabled` is why nothing is logged by default?  You would have to changed this to `workflow` from line 21 for the log to be active?

mjohnson12 (approved):

Workflow/Sources/WorkflowLogger.swift
jamieQ: that's right, this is the change that effectively turns off logging by default. consumers would have to do (at a minimum) `WorkflowLogging.enabled = true` to allow anything to be logged. i can add some additional docs to the files clarifying this

mjohnson12: 👍  Looks fine to me.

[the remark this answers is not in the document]
kyleve: Note: Recently changed this a bit, in particular making this a `private (set)` and checking `isEnabled` and bailing early from methods: https://github.com/square/Blueprint/pull/364

Workflow/Sources/WorkflowLogger.swift
```diff
@@ -14,10 +14,64 @@
...
+    /// Global setting to enable or disable logging.
+    /// Note, this is independent of the specified `config` value, and simply governs whether
+    /// the runtime should emit any logs.
+    public static var enabled: Bool {
+        get { OSLog.active === OSLog.workflow }
+        set { OSLog.active = newValue ? .workflow : .disabled }
+    }
+
+    /// Configuration options used to determine which activities are logged.
+    public static var config: Config = .debug
```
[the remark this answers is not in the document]
kyleve: I think that seems correct

Workflow/Sources/WorkflowNode.swift
```diff
@@ -74,11 +74,11 @@ final class WorkflowNode<WorkflowType: Workflow> {
         onOutput?(output)
     }
 
-    func render() -> WorkflowType.Rendering {
-        WorkflowLogger.logWorkflowStartedRendering(ref: self)
+    func render(isRootNode: Bool = false) -> WorkflowType.Rendering {
```
kyleve, line 77: Do we need to also pass this in the render tester as well?

kyleve (approved):

Workflow/Sources/WorkflowNode.swift
jamieQ: i don't think so, since that type doesn't invoke the render method of the internal representation of a node (it calls into the 'proper' render method of the client-specified Workflow).

kyleve: Ah righto

PR #153, opened 2022-08-27
Updates to make it easier to avoid using `DescribedViewController`, allowing flattening of VC hierarchies.

kyleve:
This change makes it easier to avoid using `DescribedViewController`, which adds a (largely needless) layer to view controller hierarchies. I'll leave comments inline on relevant additions with comments.

For Square folks, I recommend seeing how this is used in context in https://github.com/squareup/market/pull/4559.

## Before
image

## After
image

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowUI.podspec
+++ b/WorkflowUI.podspec
@@ -25,6 +25,10 @@ Pod::Spec.new do |s|
     test_spec.framework = 'XCTest'
     test_spec.library = 'swiftos'
     test_spec.dependency 'WorkflowReactiveSwift', "#{s.version}"
+
+    # Create an app host so that we can host
+    # view or view controller based tests in a real environment.
+    test_spec.requires_app_host = true
   end
 end
 
--- a/WorkflowUI/Sources/Container/ContainerViewController.swift
+++ b/WorkflowUI/Sources/Container/ContainerViewController.swift
@@ -27,7 +27,7 @@
             return workflowHost.output
         }
 
-        internal let rootViewController: DescribedViewController
+        private(set) var rootViewController: UIViewController
 
         private let workflowHost: WorkflowHost<RootWorkflow<ScreenType, Output>>
 
@@ -35,14 +35,18 @@
 
         public var rootViewEnvironment: ViewEnvironment {
             didSet {
-                // Re-render the current rendering with the new environment
-                render(screen: workflowHost.rendering.value, environment: rootViewEnvironment)
+                update(screen: workflowHost.rendering.value, environment: rootViewEnvironment)
             }
         }
 
         public init<W: AnyWorkflowConvertible>(workflow: W, rootViewEnvironment: ViewEnvironment = .empty) where W.Rendering == ScreenType, W.Output == Output {
             self.workflowHost = WorkflowHost(workflow: RootWorkflow(workflow))
-            self.rootViewController = DescribedViewController(screen: workflowHost.rendering.value, environment: rootViewEnvironment)
+
+            self.rootViewController = workflowHost
+                .rendering
+                .value
+                .buildViewController(in: rootViewEnvironment)
+
             self.rootViewEnvironment = rootViewEnvironment
 
             super.init(nibName: nil, bundle: nil)
@@ -56,7 +60,8 @@
                 .take(during: lifetime)
                 .observeValues { [weak self] screen in
                     guard let self = self else { return }
-                    self.render(screen: screen, environment: self.rootViewEnvironment)
+
+                    self.update(screen: screen, environment: self.rootViewEnvironment)
                 }
         }
 
@@ -69,8 +74,10 @@
             fatalError("init(coder:) has not been implemented")
         }
 
-        private func render(screen: ScreenType, environment: ViewEnvironment) {
-            rootViewController.update(screen: screen, environment: environment)
+        private func update(screen: ScreenType, environment: ViewEnvironment) {
+            update(child: \.rootViewController, with: screen, in: environment)
+
+            updatePreferredContentSizeIfNeeded()
         }
 
         override public func viewDidLoad() {
@@ -78,6 +85,7 @@
 
             view.backgroundColor = .white
 
+            rootViewController.view.frame = view.bounds
             view.addSubview(rootViewController.view)
 
             updatePreferredContentSizeIfNeeded()
--- a/WorkflowUI/Sources/Screen/Screen.swift
+++ b/WorkflowUI/Sources/Screen/Screen.swift
@@ -16,6 +16,8 @@
 
 #if canImport(UIKit)
 
+    import UIKit
+
     /// Screens are the building blocks of an interactive application.
     ///
     /// Conforming types contain any information needed to populate a screen: data,
@@ -24,6 +26,30 @@
         /// A view controller description that acts as a recipe to either build
         /// or update a previously-built view controller to match this screen.
         func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription
+    }
+
+    extension Screen {
+        /// If the given view controller is of the correct type to be updated by this screen.
+        ///
+        /// If your view controller type can change between updates, call this method before invoking `update(viewController:with:)`.
+        public func canUpdate(viewController: UIViewController, with environment: ViewEnvironment) -> Bool {
+            viewControllerDescription(environment: environment).canUpdate(viewController: viewController)
+        }
+
+        /// Update the given view controller with the content from the screen.
+        ///
+        /// ### Note
+        /// You must pass a view controller previously created by a compatible `ViewControllerDescription`
+        /// that passes `canUpdate(viewController:with:)`. Failure to do so will result in a fatal precondition.
+        public func update(viewController: UIViewController, with environment: ViewEnvironment) {
+            viewControllerDescription(environment: environment).update(viewController: viewController)
+        }
+
+        /// Construct and update a new view controller as described by this Screen.
+        /// The view controller will be updated before it is returned, so it is fully configured and prepared for display.
+        public func buildViewController(in environment: ViewEnvironment) -> UIViewController {
+            viewControllerDescription(environment: environment).buildViewController()
+        }
     }
 
 #endif
--- a/WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
+++ b/WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2020 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+
+    import UIKit
+
+    extension UpdateChildScreenViewController where Self: UIViewController {
+        /// Updates the view controller at the given `child` key path with the
+        /// `ViewControllerDescription` from `screen`. If the type of the underlying view
+        /// controller changes between update passes, this method will remove
+        /// the old view controller, create a new one, update it, and insert it into the view controller hierarchy.
+        ///
+        /// The view controller at `child` must be a child of `self`.
+        ///
+        /// - Parameters:
+        /// - parameter child: The `KeyPath` which describes what view controller to update. This view controller must be a direct child of `self`.
+        /// - parameter screen: The `Screen` instance to apply to the view controller.
+        /// - parameter environment: The `environment` to used when updating the view controller.
+        /// - parameter onChange: A callback called if the view controller instance changed.
+        ///
+        public func update<VC: UIViewController, ScreenType: Screen>(
+            child: ReferenceWritableKeyPath<Self, VC>,
+            with screen: ScreenType,
+            in environment: ViewEnvironment,
+            onChange: (VC) -> Void = { _ in }
+        ) {
+            let description = screen.viewControllerDescription(environment: environment)
+
+            let existing = self[keyPath: child]
+
+            if description.canUpdate(viewController: existing) {
+                // Easy path: Just update the existing view controller if we can do that.
+                description.update(viewController: existing)
+            } else {
+                // If we can't update the view controller, that means its type changed.
+                // We'll need to make a new view controller and swap over to it.
+
+                let old = existing
+
+                // Make the new view controller.
+
+                let new = description.buildViewController() as! VC
+
+                // We already have a reference to the old vc above, update the keypath to the new one.
+
+                self[keyPath: child] = new
+
+                // We should only add the view controller if the old one was already within the parent.
+
+                if let parent = old.parent {
+                    precondition(
+                        parent == self,
+                        """
+                        The parent of the child view controller must be \(self). Instead, it was \(parent). \
+                        Please call `update(child:)` on the correct parent view controller.
+                        """
+                    )
+
+                    // Begin the transition: Signal the new vc will begin moving in, and the old one, out.
+
+                    parent.addChild(new)
+                    old.willMove(toParent: nil)
+
+                    if
+                        parent.isViewLoaded,
+                        old.isViewLoaded,
+                        let container = old.view.superview {
+                        // We will only add the view to the hierarchy if
+                        // the parent's view is loaded, and the existing view
+                        // is loaded, and the old view was in a superview.
+
+                        // We will only perform appearance transitions if we're visible.
+
+                        let isVisible = parent.view.window != nil
+
+                        // The view should end up with the same frame.
+
+                        new.view.frame = old.view.frame
+
+                        if isVisible {
+                            new.beginAppearanceTransition(true, animated: false)
+                            old.beginAppearanceTransition(false, animated: false)
+                        }
+
+                        container.insertSubview(new.view, aboveSubview: old.view)
+                        old.view.removeFromSuperview()
+
+                        if isVisible {
+                            new.endAppearanceTransition()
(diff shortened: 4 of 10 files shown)


Workflow.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
+  s.ios.deployment_target = '12.0'
```
mjohnson12, line 16: Minor:  Should probably bump the package.swift target version as well.

kyleve: This will get reverted before the PR opens for review; it was to get around Cocoapods erroring locally on integration.

mjohnson12: 👍

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,110 @@
...
+
+                let new = description.buildViewController() as! VC
+
+                // We already have a reference to the old vc above, update the keypath to the new one.
+
+                self[keyPath: child] = new
+
+                // We should only add the view controller if the old one was already within the parent.
+
+                if let parent = old.parent {
```
kyleve, line 53: Also assert the parent is self

WorkflowUI/Tests/UIViewControllerExtensionTests.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        private struct TestScreen2: Screen {
+            func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
+                ViewControllerDescription(type: VC2.self, build: { VC2() }, update: { _ in })
+            }
+        }
+
+        func test_update() {
+            XCTFail("TODO")
```
kyleve, line 42: Will finish this up once we're OK with the direction.

Workflow.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowReactiveSwift.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowReactiveSwiftTesting.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowRxSwift.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowRxSwiftTesting.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+        ///
+        /// The view controller at `child` must be a child of `self`.
+        ///
+        /// - Parameters:
+        /// - parameter child: The `KeyPath` which describes what view controller to update. This view controller must be a direct child of `self`.
+        /// - parameter screen: The `Screen` instance to apply to the view controller.
+        /// - parameter environment: The `environment` to used when updating the view controller.
+        /// - parameter onChange: A callback called if the view controller instance changed.
+        ///
+        public func update<VC: UIViewController, ScreenType: Screen>(
```
kyleve, line 35: Useful for doing one-liner updates for contained view controllers. Eg, in Market we do:

```
private var content : UIViewController

...

override func screenDidChange(from previousScreen: ModalHostContainer, previousEnvironment: ViewEnvironment) {

    update(child: \.content, with: screen.content, in: environment)

    ...
}
```

WorkflowTesting.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowUI.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.cocoapods_version = '>= 1.7.0'
 
   s.swift_versions = ['5.0']
-  s.ios.deployment_target = '11.0'
```
kyleve, line 16: - [ ] Revert before merge (was needed to get `pod gen` working in `Market)

WorkflowUI/Sources/Screen/Screen.swift
```diff
@@ -26,4 +26,10 @@
         func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription
     }
 
+    extension Screen {
+        public func update(viewController: UIViewController, with environment: ViewEnvironment) {
```
kyleve, line 30: - [ ] Docs

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+                    new.didMove(toParent: parent)
+                    old.removeFromParent()
+                }
+
+                onChange(new)
+            }
+        }
+    }
+
+    public protocol UpdateChildScreenViewController {}
```
kyleve, line 125: This dance is needed so we can reference `Self` in the extension method.

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -118,12 +125,50 @@
...
             update(viewController)
         }
     }
 
+    extension ViewControllerDescription {
+        /// Describes the `UIViewController` type that backs the `ViewControllerDescription`
+        /// in a way that is `Equatable` and `Hashable`. When implementing view controller
+        /// updating and diffing, you can use this type to identify if the backing view controller
+        /// type changed.
+        public struct KindIdentifier: Hashable {
```
kyleve, line 141: Since this type is `Hashable`, you can use it to salt your ID types, making handling diffs easier when the underlying type of the VC in the view description changes.

mjohnson12 (approved):
This looks great.  I really like the `KindIdentifier`.

WorkflowUI/Sources/Container/ContainerViewController.swift
```diff
@@ -69,8 +75,10 @@
             fatalError("init(coder:) has not been implemented")
         }
 
-        private func render(screen: ScreenType, environment: ViewEnvironment) {
-            rootViewController.update(screen: screen, environment: environment)
+        private func update(screen: ScreenType, environment: ViewEnvironment) {
+            update(child: \.rootViewController, with: screen, in: environment)
+
+            updatePreferredContentSizeIfNeeded()
```
n8chur, line 81: Nice

WorkflowUI/Sources/Screen/Screen.swift
n8chur: Is this used?

It seems easy enough to just call the logic this abstracts, I'd be fine leaving this convenience out :shrug:

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+
+    import UIKit
+
+    extension UpdateChildScreenViewController where Self: UIViewController {
```
n8chur, line 21: Why not put this directly on `extension UIViewController` and get rid of the protocol?

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+        /// Updates the view controller at the given `child` key path with the
+        /// `ViewControllerDescription`from `screen`. If the type of the underlying view
```
n8chur, line 23:
```suggestion
        /// `ViewControllerDescription` from `screen`. If the type of the underlying view
```

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+        /// controller changes between update passes, this method will remove
+        /// the old view controller, create a new one, update it, and insert it into the view controller hierarchy.
+        ///
+        /// The view controller at `child` must be a child of `self`.
+        ///
+        /// - Parameters:
+        /// - parameter child: The `KeyPath` which describes what view controller to update. This view controller must be a direct child of `self`.
+        /// - parameter screen: The `Screen` instance to apply to the view controller.
+        /// - parameter environment: The `environment` to used when updating the view controller.
```
n8chur, line 32:
```suggestion
        /// - parameter environment: The `environment` used when updating the view controller.
```

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+
+                    // Begin the transition: Signal the new vc will begin moving in, and the old one, out.
+
+                    parent.addChild(new)
+                    old.willMove(toParent: nil)
+
+                    if
+                        parent.isViewLoaded,
+                        old.isViewLoaded,
+                        let container = old.view.superview {
```
n8chur, line 81: If these views were not loaded when this is called, how do we ensure we still `insertSubview` etc.?

WorkflowUI/Sources/ViewControllerDescription/ViewControllerDescription.swift
```diff
@@ -41,11 +41,15 @@
...
         /// (eg, a navigation stack), you usually want to set this value to `false` to avoid
         /// duplicate updates to your children if they are created in `init`.
-        public var performInitialUpdate: Bool = true
+        public var performInitialUpdate: Bool
+
+        /// Describes the `UIViewController` type that backs the `ViewControllerDescription`
+        /// in a way that is `Equatable` and `Hashable`. When implementing view controller
+        /// updating and diffing, you can use this type to identify if the backing view controller
+        /// type changed.
+        public let kind: KindIdentifier
```
n8chur, line 50: Nice

WorkflowUI/Sources/Container/ContainerViewController.swift
```diff
@@ -69,8 +75,10 @@
             fatalError("init(coder:) has not been implemented")
         }
 
-        private func render(screen: ScreenType, environment: ViewEnvironment) {
-            rootViewController.update(screen: screen, environment: environment)
+        private func update(screen: ScreenType, environment: ViewEnvironment) {
```
n8chur, line 78: Do we need to also call this from `viewDidLoad` to ensure the views get added etc.?

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,129 @@
...
+            in environment: ViewEnvironment,
+            onChange: (VC) -> Void = { _ in }
+        ) {
+            let description = screen.viewControllerDescription(environment: environment)
+
+            let existing = self[keyPath: child]
+
+            if description.canUpdate(viewController: existing) {
+                // Easy path: Just update the existing view controller if we can do that.
+                description.update(viewController: existing)
```
n8chur, line 47: Do we potentially need to add subviews etc. here if they're not already added (due to the `isViewLoaded` checks below)?

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
kyleve: This is needed because the method references `Self`, which you can't do in an extension on a concrete type like `UIViewController`. Blueprint [does the same dance, for example](https://github.com/square/Blueprint/blob/main/BlueprintUI/Sources/ViewDescription/ViewDescription.swift#L8).

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
kyleve: We don't – we assume the parent view controller will do its own setup in `loadView/viewDidLoad` when it is asked to load the view.

WorkflowUI/Sources/Container/ContainerViewController.swift
kyleve: We do not – when the container loads its view, it will add the rootVC as a child: https://github.com/square/workflow-swift/blob/main/WorkflowUI/Sources/Container/ContainerViewController.swift#L81

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
kyleve: No – the assumption is the implementation of the view controller will do this in `loadView/viewDidLoad`, as custom parent VCs are expected to do.

WorkflowUI/Sources/Screen/Screen.swift
kyleve: It's used in Market somewhat commonly to avoid having to do the description dance which saves a line of code or so, which IMO is nice, but also fine to remove it.

jamieQ (commented):
changes seem reasonable to me. curious what you think any risks or downsides to this approach may be.

WorkflowUI/Sources/Container/ContainerViewController.swift
```diff
@@ -27,22 +27,27 @@
             return workflowHost.output
         }
 
-        internal let rootViewController: DescribedViewController
+        private(set) var rootViewController: UIViewController
```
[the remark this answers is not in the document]
kyleve: Sort of: The contained VC type is now the type of the workflow's rendered screen, not a `DescribedViewController`, since we're removing a layer of nesting. However this var was always internal, so the only way it'll affect folks is if they're reaching upwards from the contained VC, or downwards into the Container, in a non-resilient way.

WorkflowUI/Sources/Container/ContainerViewController.swift
kyleve: If we're swapping out the inner view controller impl, we need to re-ask for its size. The fact we don't do this in `DescribedViewController` seems like a bug; I can fix here.

- [ ] Fix `DescribedViewController`

WorkflowUI/Sources/Container/ContainerViewController.swift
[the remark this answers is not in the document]
jamieQ: i apologize, i should have been more specific. i was wondering about the access control changes & `let` vs `var` in particular. this didn't look like it was being set anywhere but init, so wondering if this could just be `private let`, or maybe just `let` to keep test visibility.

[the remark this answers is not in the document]
kyleve: It is set in the update method, via `update(child: \.rootViewController, with: screen, in: environment)` (the key path) if the underlying VC type has to change.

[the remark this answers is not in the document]
jamieQ: aah, right it was a writable keypath. thanks for clarifying!

WorkflowUI/Tests/UIViewControllerExtensionTests.swift
```diff
@@ -0,0 +1,270 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+
+    import XCTest
+    @testable import WorkflowUI
+
+    class UIViewControllerExtensionTests: XCTestCase {
```
kyleve, line 22: Tests have been added

WorkflowUI/Sources/ViewControllerDescription/UIViewController+Extensions.swift
kyleve: Fixing locally

WorkflowUI/Tests/ViewControllerDescriptionTests.swift
```diff
@@ -148,4 +148,23 @@
         }
     }
 
+    class ViewControllerDescription_KindIdentifierTests: XCTestCase {
+        private final class VC1: UIViewController {}
+        private final class VC2: UIViewController {}
+
+        func test_kind() {
```
jamieQ, line 155: what's the expected behavior if you had, say, `class VC3: VC2`?

kyleve: Same as before; whatever `is` returns for the given types.

PR #154, opened 2022-08-30
Added AsyncStreamWorker

mjohnson12:
## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowConcurrency/Sources/AsyncStreamWorker.swift
+++ b/WorkflowConcurrency/Sources/AsyncStreamWorker.swift
@@ -1,0 +1,71 @@
+//
+//  AsyncStreamWorker.swift
+//  WorkflowConcurrency
+//
+//  Created by Mark Johnson on 5/19/22.
+//
+
+import Foundation
+import Workflow
+
+/// Workers define a unit of asynchronous work.
+///
+/// During a render pass, a workflow can ask the context to await the result of a worker.
+///
+/// When this occurs, the context checks to see if there is already a running worker of the same type.
+/// If there is, and if the workers are 'equivalent', the context leaves the existing worker running.
+///
+/// If there is not an existing worker of this type, the context will kick off the new worker (via `run`).
+@available(iOS 13.0, macOS 10.15, *)
+public protocol AsyncStreamWorker: AnyWorkflowConvertible where Rendering == Void {
+    /// The type of output events returned by this worker.
+    associatedtype Output
+
+    /// Returns an AsyncStream to execute the work represented by this worker.
+    func run() -> AsyncStream<Output>
+    /// Returns `true` if the other worker should be considered equivalent to `self`. Equivalence should take into
+    /// account whatever data is meaningful to the task. For example, a worker that loads a user account from a server
+    /// would not be equivalent to another worker with a different user ID.
+    func isEquivalent(to otherWorker: Self) -> Bool
+}
+
+@available(iOS 13.0, macOS 10.15, *)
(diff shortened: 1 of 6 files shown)


WorkflowConcurrency/Sources/AsyncStreamWorker.swift
```diff
@@ -0,0 +1,71 @@
...
+            state = UUID()
+        }
+    }
+
+    func render(state: State, context: RenderContext<AsyncStreamWorkerWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: "") { lifetime in
+            let task = Task {
+                for await output in worker.run() {
+                    DispatchQueue.main.async {
```
JaviSoto, line 60: This seems a bit smelly to me. If you extract the code inside the `for await` into a `func` and tag it with `@MainActor`, will this work automatically?

WorkflowConcurrency/Sources/AsyncStreamWorker.swift
```diff
@@ -0,0 +1,71 @@
...
+                        sink.send(AnyWorkflowAction(sendingOutput: output))
+                    }
+                }
+            }
+
+            lifetime.onEnded {
+                task.cancel()
```
JaviSoto, line 67: 👍

JaviSoto (commented):
Looks good to me, although I still have limited experience with Workflow. It would be nice to see an example using this, maybe as part of unit tests?
Just had 1 comment about the dispatch call, but otherwies 👍

WorkflowConcurrency/Sources/AsyncStreamWorker.swift
mjohnson12: Done

mjohnson12:
> Looks good to me, although I still have limited experience with Workflow. It would be nice to see an example using this, maybe as part of unit tests? Just had 1 comment about the dispatch call, but otherwies 👍

Yes.   I am definitely planning on adding tests once I move from this draft to a final PR.

WorkflowConcurrency/Sources/AsyncStreamWorker.swift
```diff
@@ -0,0 +1,74 @@
...
+            state = UUID()
+        }
+    }
+
+    func render(state: State, context: RenderContext<AsyncStreamWorkerWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: "") { lifetime in
+            let task = Task {
+                for await output in worker.run() {
+                    await sendAction(output: output, sink: sink)
```
jamieQ, line 60: i haven't really validated any existing behavior of the other Worker implementations, but i do wonder if we have to be careful about using the `sink` here in any way. like if this were to suspend on L60 but by the time `sendAction` ends up running on the main thread, the Task has been canceled, will we run into problems when trying to send things through the sink?

mjohnson12: The sink isn't going to loose scope since it's retained here and the workflow here is the internal workflow used when you run the worker.  I don't think it's an issue but I was going to test the Task being cancelled anyways so I can try it.

mjohnson12: I've tried different scenarios of forcing the Task to be cancelled and upon cancellation it doesn't attempt to call sink again after cancelling.

PR #155, opened 2022-09-02
[gardening]: logging utility method cleanup

jamieQ:
### Description

- clean up internal logging utility method

### DIFF
--- a/Workflow/Sources/WorkflowLogger.swift
+++ b/Workflow/Sources/WorkflowLogger.swift
@@ -140,8 +140,7 @@ final class WorkflowLogger {
         isRootNode: Bool
     ) {
         if #available(iOS 12.0, macOS 10.14, *) {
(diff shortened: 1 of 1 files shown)


Workflow/Sources/WorkflowLogger.swift
```diff
@@ -174,16 +172,15 @@ final class WorkflowLogger {
 
     // MARK: - Utilities
 
-    private static func shouldLogRenderTimingsForMode(
-        _ renderLoggingMode: WorkflowLogging.Config.RenderLoggingMode,
```
jamieQ, line 178: we're just using the static config value, so dropped the unused parameter.

reidmain (approved):

Kiesco08 (approved):

kyleve (approved):

PR #160, opened 2022-10-03
Added Concurrency Testing

mjohnson12:
## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowCombine/TestingTests/PublisherTests.swift
+++ b/WorkflowCombine/TestingTests/PublisherTests.swift
@@ -12,6 +12,7 @@ import WorkflowTesting
 import XCTest
 @testable import WorkflowCombineTesting
 
+@available(iOS 13.0, macOS 10.15, *)
 class PublisherTests: XCTestCase {
     func testPublisherWorkflow() {
         TestWorkflow()
--- a/WorkflowConcurrency/Testing/TaskTesting.swift
+++ b/WorkflowConcurrency/Testing/TaskTesting.swift
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2022 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if DEBUG
+
+    import Combine
+    import Workflow
+    import WorkflowTesting
+    import XCTest
+    @testable import WorkflowConcurrency
+
+    @available(macOS 10.15, *)
+    @available(iOS 13.0, *)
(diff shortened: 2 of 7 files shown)


WorkflowCombine/TestingTests/PublisherTests.swift
```diff
@@ -12,6 +12,7 @@ import WorkflowTesting
 import XCTest
 @testable import WorkflowCombineTesting
 
+@available(iOS 13.0, macOS 10.15, *)
```
jamieQ, line 15: why were all these availability annotations needed?

Development.podspec
```diff
@@ -181,4 +181,12 @@ Pod::Spec.new do |s|
...
   #   test_spec.dependency 'WorkflowTesting'
   # end
+  
+  # s.test_spec 'WorkflowConcurrencyTestingTests' do |test_spec|
+  #   test_spec.requires_app_host = true
+  #   test_spec.source_files = 'WorkflowConcurrency/TestingTests/**/*.swift'
+  #   test_spec.framework = 'XCTest'
+  #   test_spec.dependency 'WorkflowTesting'
+  #   test_spec.dependency 'WorkflowConcurrencyTesting'
+  # end
```
Kiesco08, line 191: Do we need to add this if it's disabled?

WorkflowConcurrency/Testing/TaskTesting.swift
```diff
@@ -0,0 +1,50 @@
...
+    @available(iOS 13.0, *)
+    extension RenderTester {
+        /// Expect a `Task`s.
+        ///
+        /// `TaskWorkflow` is used to execute to `Task`s.
+        ///
+        /// - Parameters:
+        ///   - producingOutput: An output that should be returned when this worker is requested, if any.
+        ///   - key: Key to expect this `Workflow` to be rendered with.
+        public func expectTask<Value>(
```
jamieQ, line 35: nit: are there other examples of `expectXYZ(xyz: ...)` in these utility methods? the ones i easily found all looked like `expect(xyz: ...)` and i'm inclined to keep this the same to preserve 'muscle memory' for people if that's the only existing naming convention.

WorkflowCombine/TestingTests/PublisherTests.swift
mjohnson12: I added the testing target for those to the package and because the package is iOS 12+ right now.  You can't set minimum OS targets for products only packages unfortunately.

Development.podspec
mjohnson12: It's added so when we switch the minimum OS to higher than 12 we just uncomment so we don't forget to add that target.

WorkflowConcurrency/Testing/TaskTesting.swift
mjohnson12: WorkflowConcurrency follows the same patterns as WorkflowCombine.  It's `expectPublisher` in `WorkflowCombineTesting`.  I'm fine with changing both of these.

Kiesco08 (approved):

WorkflowConcurrency/Testing/TaskTesting.swift
jamieQ: okay cool, i think that'd be a nice marginal improvement. good to keep logically equivalent APIs using the same naming conventions IMO.

jamieQ (approved):
lgtm. just had one thought about unifying naming conventions

PR #167, opened 2022-12-12
Add StateMutationSink

n8chur:
## Background

From @dhavalshreyas's [proposal in Notion](https://www.notion.so/sellermobile/Workflow-Boilerplate-Reduction-be0ccb80c5a044c0b564d09ddd1ff63f#62d55b4121aa428095b7370ebce43c9e) (with slightly updated examples):

`Sink`s are used to process `Action`s, either from `SideEffect`s or from user-interaction events.

We currently have `makeSink` to process any `Action` and `makeOutputSink`, which provides a `Sink` that sends the `Output`.

The proposal here is to introduce `makeStateMutationSink`, which would provide a `Sink` that automatically mutates `State` without an intermediate `Action`.

**Before:**

```swift
struct MyWorkflow: Workflow {
    enum Action: WorkflowAction {
        typealias WorkflowType = MyWorkflow

        case updateTapCount(Int)

        func apply(toState state: inout Int) -> MyWorkflow.Output? {
            switch self {
            case .updateTapCount(let count):
                state.tapCount = count
                return nil
            }
            return nil
        }
    }

    func render(state: State, context: RenderContext<Self>) -> Rendering {
        let sink = context.makeSink(of: Action.self)
        return Screen(onTap: { _ in
            sink.send(.updateTapCount(state.tapCount + 1))
        })
    }
}
```

**After:**

```swift
struct MyWorkflow: Workflow {
    func render(state: State, context: RenderContext<Self>) -> Rendering {
        let stateMutationSink = context.makeStateMutationSink()
        return Screen(onTap: { _ in
            stateMutationSink.send(\.tapCount, value: state.tapCount + 1)
        })
    }
}
```

---

This convenience has been baking in `ios-register`'s `WorkflowExperimental` framework for quite some time now and has proven useful to many.

## Additions to existing convenience

These changes area a direct port of that code (including the associated tests) with one additional convenience which would allow for mutating state with a closure rather than just the `KeyPath`/`Value` syntax:

```swift
struct MyWorkflow: Workflow {
    func render(state: State, context: RenderContext<Self>) -> Rendering {
        let stateMutationSink = context.makeStateMutationSink()
        return Screen(onTap: { _ in
            stateMutationSink.send { state in
               state.tapCount += 1
               state.tapped = true
            }
        })
    }
}
```

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Workflow/Tests/StateMutationSinkTests.swift
+++ b/Workflow/Tests/StateMutationSinkTests.swift
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2022 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import ReactiveSwift
+import Workflow
+import XCTest
+
+final class StateMutationSinkTests: XCTestCase {
+    var output: Signal<Int, Never>!
+    var input: Signal<Int, Never>.Observer!
+
+    override func setUp() {
+        (output, input) = Signal<Int, Never>.pipe()
+    }
+
+    func test_initialValue() {
+        let host = WorkflowHost(workflow: TestWorkflow(value: 100, signal: output))
+        XCTAssertEqual(0, host.rendering.value)
+    }
+
+    func test_singleUpdate() {
+        let host = WorkflowHost(workflow: TestWorkflow(value: 100, signal: output))
+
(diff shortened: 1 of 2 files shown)


Workflow/Tests/StateMutationSinkTests.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    func test_initialValue() {
+        let host = WorkflowHost(workflow: TestWorkflow(value: 100, signal: output))
+        XCTAssertEqual(0, host.rendering.value)
+    }
+
+    func test_singleUpdate() {
+        let host = WorkflowHost(workflow: TestWorkflow(value: 100, signal: output))
+
+        let gotValueExpectation = expectation(description: "Got expected value")
```
jamieQ, line 37: this doesn't really affect the semantics of the tests at all, but i _think_ the expectation machinery isn't strictly needed in this instance. the processing should happen synchronously as soon as the values are received.

jamieQ (approved):

Workflow/Tests/StateMutationSinkTests.swift
n8chur: Would you prefer that I try removing it?

I just directly ported the test from `ios-register` and didn't dig into how I might be able to improve these tests much...

PR #168, opened 2022-12-13
[feat]: add runtime observation API

jamieQ:
## Issue

the Swift implementation of the Workflow library currently does not offer much support for consumers to tie into runtime operations in which they may be interested. by contrast, the Kotlin implementation has supported various facilities for this over time, and currently allows most runtime methods to be observed or [even replaced](https://github.com/square/workflow-kotlin/blob/a39c2883708194c157ba592d2982594a86d6f341/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt#L19-L22) by clients.

## Description

- adds a new type `WorkflowSession` that represents metadata about the node backing a particular `Workflow` 'instance' in the tree. this was largely modeled after the [analogous concept](https://github.com/square/workflow-kotlin/blob/d39fe73844d19f316509390e17d381f56f17557a/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt#L114) in the Kotlin library. this type includes:
  1. the type of the corresponding workflow
  1. the string key used to render the workflow
  1. an opaque identifier type (currently implemented via an underlying increasing global counter)
  1. an optional reference to the parent node's session data
- adds the `WorkflowObserver` protocol, which provides (optional) runtime hooks for the following:
  1. session start/end (i.e. when a node is installed/removed from the tree)
  1. when a Worfklow creates its initial state
  1. before/after a Worfklow is rendered
  1. when an existing Workflow is updated
  1. when a WorkflowAction is received
  1. before/after a WorkflowAction is applied to a Workflow's State
- provides an internal `WorkflowObserver` implementation to multiplex observers provided by clients
- adds/updates tests

## Checklist

- [x] Unit Tests
- [x] ~UI Tests~ N/A
- [x] ~Snapshot Tests (iOS only)~ N/A


Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,262 @@
...
+    /// Indicates the start of a `WorkflowSession`, which tracks the life of the underlying `WorkflowNode` used to provide renderings for a given `Workflow` type and rendering key.
+    /// - Parameter session: The `WorkflowSession` that began.
+    func sessionDidBegin(
+        _ session: WorkflowSession
+    )
+
+    /// Marks the end of a `WorkflowSession`, indicating that the corresponding `WorkflowNode` has been removed from the tree of Workflows.
+    /// - Parameter session: The `WorkflowSession` that ended.
+    func sessionDidEnd(
+        _ session: WorkflowSession
```
amorde, line 33: it looks like a few of these don't have access to `state`, and none of them have access to the underlying workflow instance. is there an underlying limitation preventing that? I somewhat assumed `WorkflowSession` would contain some of those things at first glance but realized I was wrong after further reading

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,262 @@
...
+        state: WorkflowType.State,
+        session: WorkflowSession
+    )
+
+    /// Called after a `WorkflowAction` is received, but **before** it has been handled and propagated up the tree.
+    /// - Parameters:
+    ///   - action: The action that was received.
+    ///   - session: The `WorkflowSession` corresponding to the backing `WorkflowNode`.
+    func workflowDidReceiveAction<Action: WorkflowAction>(
+        _ action: Action,
```
amorde, line 77: can the workflow itself be passed in here, and its state?

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,262 @@
...
+        private static var _nextRawID: UInt64 = 0
+        private static func _makeNextSessionID() -> UInt64 {
+            _nextRawID += 1
+            return _nextRawID
+        }
+
+        let rawIdentifier: UInt64 = Self._makeNextSessionID()
+    }
+
+    private indirect enum IndirectParent {
```
amorde, line 101: out of curiosity, what drove the need for this?

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,262 @@
...
+        let callbacks = observers.compactMap {
+            $0.workflowWillRender(workflow, state: state, session: session)
+        }
+
+        guard !callbacks.isEmpty else {
+            return nil
+        }
+
+        return { rendering in
+            for callback in callbacks.reversed() {
```
amorde, line 221: `workflowWillRender` is called in order, but the callbacks are called in reverse order. was that intentional?

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,262 @@
...
+            return nil
+        }
+    }
+
+    init<WorkflowType: Workflow>(
+        workflow: WorkflowType,
+        renderKey: String,
+        parent: WorkflowSession?
+    ) {
+        self.workflowType = WorkflowType.self
```
amorde, line 136: any reason for not storing the workflow itself?

Workflow/Sources/WorkflowObserver.swift
jamieQ: the session stuff in particular was largely modeled on the analogous concepts/interfaces in the [kotlin repo](https://github.com/square/workflow-kotlin/blob/a39c2883708194c157ba592d2982594a86d6f341/workflow-runtime/src/commonMain/kotlin/com/squareup/workflow1/WorkflowInterceptor.kt#L64-L67).

i think i need a bit of clarification on this thought:

> none of them have access to the underlying workflow instance

by 'underlying workflow instance', what do you mean exactly? in many of them the type conforming to `Workflow` is passed as a parameter for the observer to use.

> is there an underlying limitation preventing that?

for the session lifecycle in particular, the state and corresponding workflow (conforming type instance) could be passed. though there is a question about how to order the `sessionDidBegin` and `workflowDidMakeInitialState`.

Workflow/Sources/WorkflowObserver.swift
jamieQ: it probably can – i think i had experimented with some variants of that, and seem to recall some issues that weren't straightforward to resolve, but i will take a look at it again.

Workflow/Sources/WorkflowObserver.swift
jamieQ: structs and enums cannot contain stored properties that reference their type. originally i had this as a class, but swapped to the 'indirect enum case' (i.e. a custom optional implementation that allows recursion), mostly just to see if implementing this with the value types was possible.

IMO it is a bit unusual, but seems to work. though it might also make sense to remove the parent reference entirely if it doesn't seem valuable. it does let you walk from a given node up to the root though, which might be useful in some circumstances.

Workflow/Sources/WorkflowObserver.swift
jamieQ: yeah, maybe that's weird (also the behavior should be documented if we keep it). the motivation was the performance tracing use case – you want the start time to be recorded first, and the end time to happen after everything else is done, including any other observers running.

Workflow/Sources/WorkflowObserver.swift
jamieQ: i was thinking of the session as like mostly fixed metadata about the 'node' that persists over time. the actual `Workflow` instance (that clients provide) that is associated with a given node can change over time, so if we stored that here we'd have to deal with that fact. also, this is largely modeled from the Kotlin implementation, which doesn't store the Workflow directly on this type (or, the 'props' as they're called in that context).

mjohnson12 (approved):
Looks great to me

Workflow/Sources/WorkflowObserver.swift
amorde: ah yep that makes sense!

Workflow/Sources/WorkflowHost.swift
```diff
@@ -49,10 +49,18 @@ public final class WorkflowHost<WorkflowType: Workflow> {
     /// - Parameter workflow: The root workflow in the hierarchy
     /// - Parameter debugger: An optional debugger. If provided, the host will notify the debugger of updates
```
square-tomb, line 50: Should we add a line here to document `observers`?

Workflow/Sources/WorkflowHost.swift
```diff
@@ -49,10 +49,18 @@ public final class WorkflowHost<WorkflowType: Workflow> {
...
     ///                       to the workflow hierarchy as state transitions occur.
-    public init(workflow: WorkflowType, debugger: WorkflowDebugger? = nil) {
+    public init(
+        workflow: WorkflowType,
+        observers: [WorkflowObserver]? = nil,
```
square-tomb, line 54: Do `[]` and `nil` have different meaning as values for the `observers` parameter?

Workflow/Sources/WorkflowNode.swift
```diff
@@ -82,17 +122,29 @@ final class WorkflowNode<WorkflowType: Workflow> {
...
+        let renderObserverCompletion = observer?.workflowWillRender(
+            workflow,
+            state: state,
+            session: session
+        )
+
+        let rendering: WorkflowType.Rendering
+
         defer {
+            renderObserverCompletion?(rendering)
```
square-tomb, line 134: I haven't seen this define-defer-inititalize pattern before—nice!

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,322 @@
...
+/// - The `String` key used when the workflow was rendered.
+/// - An `Identifier` type that is unique across program execution.
+/// - An optional reference to a parent `WorkflowSession`, to differentiate root nodes.
+public struct WorkflowSession {
+    public struct Identifier: Hashable {
+        private static var _nextRawID: UInt64 = 0
+        private static func _makeNextSessionID() -> UInt64 {
+            _nextRawID += 1
+            return _nextRawID
+        }
```
square-tomb, line 111: We might want to mention in the docstring that `WorkflowSession.init()` is not thread-safe.

Workflow/Sources/WorkflowObserver.swift
square-tomb: To make sure I understand, if you deleted the `indirect` keyword from the declaration of `IndirectParent`, it wouldn't compile?

I see that is true if some `case` of `IndirectParent` has an associated value of type `IndirectParent`, but I didn't realize that's true also if the associated value type is only _mutually_ recursive with `IndirectParent`, like `WorkflowSession` is.

square-tomb (approved):

Workflow/Tests/WorkflowObserverTests.swift
```diff
@@ -0,0 +1,537 @@
...
+        let rendering = node.render()
+        node.enableEvents()
+
+        XCTAssertEqual(willApplyActions, [])
+        XCTAssertEqual(didApplyActions, [])
+
+        rendering.toggle()
+
+        XCTAssertEqual(willApplyState, false)
+        XCTAssertEqual(didApplyState, true)
```
JaviSoto, line 197: `XCTAssertTrue/False`

Workflow/Tests/WorkflowObserverTests.swift
```diff
@@ -0,0 +1,537 @@
...
+
+            let observer = host.rootNode.observer as? ChainedWorkflowObserver
+            XCTAssertNotNil(observer)
+
+            XCTAssertEqual(observer?.observers.count, 2)
+            XCTAssert(observer?.observers.first is TestObserver)
+            XCTAssert(observer?.observers.last is NoOpObserver)
+        }
+    }
+}
```
JaviSoto, line 451: These are super thorough!! 👍👍👍

Workflow/Tests/WorkflowObserverTests.swift
```diff
@@ -0,0 +1,537 @@
...
+
+// MARK: - Utilities
+
+private final class TestObserver: WorkflowObserver {
+    var onSessionBegan: ((WorkflowSession) -> Void)?
+    var onSessionEnded: ((WorkflowSession) -> Void)?
+    /// (Workflow, State, Session) -> Void
+    var onDidMakeInitialState: ((Any, Any, WorkflowSession) -> Void)?
```
JaviSoto, line 459: What do you think about making this class generic over `WorkflowType`?

JaviSoto (commented):
This was super easy to follow, the approach makes a ton of sense, and the tests are fantastic. This is really great!

Workflow/Sources/WorkflowObserver.swift
jamieQ: > To make sure I understand, if you deleted the indirect keyword from the declaration of IndirectParent, it wouldn't compile?

@square-tomb that's right. if i remove it, i get the following compiler error at the declaration of the `_indirectParent` property:

> Value type 'WorkflowSession' cannot have a stored property that recursively contains it

i assume this is because value types are to be stack-allocated, and so their entire layout must have a fixed size (or at least a bounded size?) so the compiler knows how to set up the function prologues appropriately. if a value type contains a recursive reference (anywhere in a descendant property), then you get this error, presumably because the type's layout size cannot be determined.

> Another typical workaround is to use a "box"

@JaviSoto yep, and i believe that is essentially what's done internally when you mark an enum as `indirect`. i found [this thread](https://forums.swift.org/t/who-benefits-from-the-indirect-keyword/20167) had some useful information and links on the topic.

Workflow/Sources/WorkflowHost.swift
jamieQ: no – i can remove the optionality

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,322 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+// MARK: - WorkflowObserver
+
+/// The `WorkflowObserver` protocol provides an interface to receive updates during the runtime's
+/// execution loop. All requirements are optional, and no-ops are provided by default.
```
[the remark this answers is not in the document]
jamieQ: Xcode 14 appears to provide a somewhat-improved check for simple typos – it will warn if an instance method 'nearly matches' a defaulted requirement. some fairly straightforward unit tests of any conformance would likely suffice as defense against the typo concern, i think.

i'm not convinced that if you adopt this protocol you'd inherently want to observe everything. for example, if you wanted to implement separate observers to handle different responsibilities (e.g. one for measuring render times, one for recording action emissions, etc), then you'd have a fair amount of boilerplate to provide no-op implementations of the irrelevant bits. i'll think on it a bit more though...

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,322 @@
...
+    func sessionDidEnd(
+        _ session: WorkflowSession
+    )
+
+    /// Indicates a `Workflow` produced its initial state value.
+    /// - Parameters:
+    ///   - workflow: The `Workflow` that just produced its initial state.
+    ///   - initialState: The `State` that was created.
+    ///   - session: The `WorkflowSession` corresponding to the backing `WorkflowNode`
+    func workflowDidMakeInitialState<WorkflowType: Workflow>(
```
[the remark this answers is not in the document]
jamieQ: no, an observer observes the runtime of an entire Workflow 'tree', which can be comprised of an arbitrary collection of Workflow types.

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,322 @@
...
+/// In particular, it captures:
+/// - The type of the corresponding `Workflow` conformance.
+/// - The `String` key used when the workflow was rendered.
+/// - An `Identifier` type that is unique across program execution.
+/// - An optional reference to a parent `WorkflowSession`, to differentiate root nodes.
+public struct WorkflowSession {
+    public struct Identifier: Hashable {
+        private static var _nextRawID: UInt64 = 0
+        private static func _makeNextSessionID() -> UInt64 {
+            _nextRawID += 1
```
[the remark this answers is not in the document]
jamieQ: i don't think overflowing would be a concern for several lifetimes, but good to know that API exists!

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,322 @@
...
+            switch parent {
+            case .some(let value):
+                self = .some(value)
+            case .none:
+                self = .none
+            }
+        }
+    }
+
+    public let workflowType: Any.Type
```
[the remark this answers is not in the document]
jamieQ: perhaps, though i'm not sure what additional utility that would provide. do you have something in mind that we would do with that info (aside from replace this property with the generic type)?

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,322 @@
...
+
+/// A `WorkflowSession`encapsulates the information that gives a `WorkflowNode` its identity.
+/// In particular, it captures:
+/// - The type of the corresponding `Workflow` conformance.
+/// - The `String` key used when the workflow was rendered.
+/// - An `Identifier` type that is unique across program execution.
+/// - An optional reference to a parent `WorkflowSession`, to differentiate root nodes.
+public struct WorkflowSession {
+    public struct Identifier: Hashable {
+        private static var _nextRawID: UInt64 = 0
```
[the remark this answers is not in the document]
jamieQ: good point! will return the existing value before incrementing

Workflow/Tests/WorkflowObserverTests.swift
```diff
@@ -0,0 +1,537 @@
...
+
+        var sessionEndedCount = 0
+        observer.onSessionEnded = { session in
+            endedSession = session
+            sessionEndedCount += 1
+        }
+
+        // session end happens in deinit, so need control over release timing
+        autoreleasepool {
+            _ = WorkflowHost(
```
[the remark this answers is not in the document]
jamieQ: reasonable point, but as written, that induces a compiler warning, and IMO the additional syntactic noise needed to achieve this isn't particularly valuable

Workflow/Tests/WorkflowObserverTests.swift
jamieQ: the values are optional, so IMO this reads better

Workflow/Tests/WorkflowObserverTests.swift
jamieQ: i'm not sure if that would help anything, since the callbacks can run on heterogenous Workflow types. i agree this doesn't seem like the most elegant approach, but it seems fairly useful as a testing device.

Workflow/Tests/WorkflowObserverTests.swift
```diff
@@ -0,0 +1,537 @@
...
+            XCTAssertEqual(observer?.observers.count, 2)
+            XCTAssert(observer?.observers.first is TestObserver)
+            XCTAssert(observer?.observers.last is NoOpObserver)
+        }
+    }
+}
+
+// MARK: - Utilities
+
+private final class TestObserver: WorkflowObserver {
```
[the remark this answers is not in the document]
jamieQ: that's true. i think i had something like that at one point, but decided against it since some of the type checking assertions then ended up being done by string comparisons which felt kinda awkward/unwieldy (or at least that's the approach i came up with at the time).

Workflow/Sources/WorkflowAction.swift
```diff
@@ -37,6 +37,10 @@ public protocol WorkflowAction {
 public struct AnyWorkflowAction<WorkflowType: Workflow>: WorkflowAction {
     private let _apply: (inout WorkflowType.State) -> WorkflowType.Output?
 
+    /// Underlying type-erased `WorkflowAction` value, if it exists. Will be nil if the
+    /// action is defined by a closure. Primarily used for testing purposes.
+    let _wrappedValue: Any?
```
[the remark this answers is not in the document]
jbmorgan: Square is on Xcode 14.2 now. Can we make this change?

Workflow/Sources/WorkflowHost.swift
```diff
@@ -36,7 +36,7 @@ public final class WorkflowHost<WorkflowType: Workflow> {
 
     private let (outputEvent, outputEventObserver) = Signal<WorkflowType.Output, Never>.pipe()
 
-    private let rootNode: WorkflowNode<WorkflowType>
+    let rootNode: WorkflowNode<WorkflowType>
```
jbmorgan, line 39: You mentioned concerns about potentially leaking some Workflow internals that could lead to Workflows hanging around longer than expected. Is this one of those places?

Workflow/Sources/WorkflowObserver.swift
```diff
@@ -0,0 +1,367 @@
...
+        switch _indirectParent {
+        case .some(let parent):
+            return parent
+        case .none:
+            return nil
+        }
+    }
+
+    /// Creates a new `WorkflowSession` instance. Note, construction of this type
+    /// is not safe to perform concurrently with respect to other instance initialization.
```
jbmorgan, line 150: Why is that? Should we be doing something to guard against misusage?

jamieQ: internally it uses a global variable to generate the unique ids which has no synchronization. i suppose we could add something more, though i'm not sure what in particular would be of value. in general i think of Workflow as a main-thread-only API at the moment (the runtime enforces this in various places), so IMO somehow instantiating this off the main thread would already be somewhat unexpected.

Workflow/Sources/WorkflowAction.swift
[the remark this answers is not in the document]
mjohnson12: Workflow is an open source repo and currently supports iOS 11+.  Making it swift 5.7+ required is definitely a breaking change.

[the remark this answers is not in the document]
JaviSoto: Ah, I kept forgetting that 👍 
Alternatively, you could do it with `#if compiler()`, but I wouldn't recommend it unless we have CI set up to make sure everything builds under both Xcode 14.0 and 14.2

Workflow/Sources/WorkflowObserver.swift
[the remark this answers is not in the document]
JaviSoto: OK that's fair 👍 (sorry I totally missed your responses before, I need to figure out my email filters…)

Workflow/Sources/WorkflowAction.swift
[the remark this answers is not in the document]
kyleve: (Not to backseat code here; but shall we do the deployment target bump some time soon? It feels fine for folks still needing < 14 to just use an old version)

Workflow/Sources/WorkflowObserver.swift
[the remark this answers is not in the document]
JaviSoto: Only useful if the observer will want to access properties in the workflow itself, but if this is going to be used just for like printing the type name then yeah making it generic isn't that useful

Workflow/Tests/WorkflowObserverTests.swift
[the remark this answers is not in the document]
JaviSoto: You can workaround the compiler warning with:

```swift
weak var workflowHost: WorkflowHost?

autoreleasepool {
  withExtendedLifetime(WorkflowHost(…)) {
    workflowHost = $0
  }
}

XCTAssertNil(workflowHost)
```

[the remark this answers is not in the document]
JaviSoto: I think making it easier to understand why a test failed is valuable. It also makes the test a bit more explicit: it makes it clearer that we're expecting that host to be deallocated by that point

Workflow/Tests/WorkflowObserverTests.swift
JaviSoto: Ah! I missed that they were optional

Workflow/Tests/WorkflowObserverTests.swift
JaviSoto: >the callbacks can run on heterogenous Workflow types

Yeah that's what I missed all around 👍

Workflow/Sources/WorkflowHost.swift
JaviSoto: Sometimes I use this little trick to make it clearer why something that one may think should be private isn't, and also to emphasize that while it isn't private, it's also no public:

```swift
// @testable
internal let rootNode: WorkflowNode<WorkflowType>
```

kyleve: You could also leverage `@_spi` here to force someone to opt into the SPI interface to access this.

jbmorgan (approved):

PR #169, opened 2022-12-14
[chore]: split WorkflowConcurrency versioning to address source-breaking change

jamieQ:
### Issue

- #164 removed a public API in the WorkflowConcurrency library, so it needs special-casing to conform to semantic versioning standards
- once we bump Workflow to 2.x, we should be able to revert to a shared version for all the utility libraries

### Description

- add WORKFLOW_CONCURRENCY_VERSION support, and bump to `2.0.0`
- update supporting files where needed


mjohnson12 (approved):
Looks good.  Should the workflow minor release change though because of the additions to the API that Westin added?

jamieQ:
> Looks good. Should the workflow minor release change though because of the additions to the API that Westin added?

yep! i will do that in a separate PR before making the new release

PR #180, opened 2023-01-16
[fix]: minor storage lifetime optimization

jamieQ:
### Issue

- to support the `mapRendering` & `mapOutput` API, `AnyWorkflow` internally stores these transformations in a storage type
- the existing method implementations would retain the existing storage object unnecessarily, when only the existing transform need be retained

### Description

- removed the unnecessary captures of the full internal storage object when mapping outputs/renderings

## Checklist

- [x] Unit Tests (existing tests suffice)
- [x] ~UI Tests~ n/a
- [x] ~Snapshot Tests (iOS only)~ n/a
- [x] ~I have made corresponding changes to the documentation~ n/a

### DIFF
--- a/Workflow/Sources/AnyWorkflow.swift
+++ b/Workflow/Sources/AnyWorkflow.swift
@@ -131,14 +131,18 @@ extension AnyWorkflow {
             return AnyWorkflow<Rendering, NewOutput>.Storage<T>(
                 workflow: workflow,
                 renderingTransform: renderingTransform,
-                outputTransform: { transform(self.outputTransform($0)) }
+                outputTransform: { [outputTransform] in
+                    transform(outputTransform($0))
+                }
             )
         }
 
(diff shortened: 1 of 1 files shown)


Workflow/Sources/AnyWorkflow.swift
```diff
@@ -131,14 +131,18 @@ extension AnyWorkflow {
             return AnyWorkflow<Rendering, NewOutput>.Storage<T>(
                 workflow: workflow,
                 renderingTransform: renderingTransform,
-                outputTransform: { transform(self.outputTransform($0)) }
+                outputTransform: { [outputTransform] in
+                    transform(outputTransform($0))
+                }
```
jamieQ, line 136: in the common use case, you have things like:

```swift
MyWorkflow()
  .asAnyWorkflow()  // creates internal storage
  .mapOutput { ... }  // creates new internal storage & would previously hold a reference to old storage
  .rendered(...)
```

now we won't keep the storage instances around longer than necessary.

square-tomb (approved):
Great catch!

PR #181, opened 2023-01-19
[feat]: add primary associated types to `Workflow` protocol

jamieQ:
### Issue

- Workflow doesn't yet expose primary associated types, which makes some forms of type-erasure less useful or impossible

### Description

- adds `Rendering` and `Output` as the primary associated types of the `Workflow` protocol

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Workflow/Sources/Workflow.swift
+++ b/Workflow/Sources/Workflow.swift
@@ -49,6 +49,39 @@
 /// }
 /// ```
 ///
+#if swift(>=5.7)
+public protocol Workflow<Rendering, Output>: AnyWorkflowConvertible {
+    /// Defines the state that is managed by this workflow.
+    associatedtype State
+
+    /// `Output` defines the type that can be emitted as output events.
+    associatedtype Output = Never
+
+    /// `Rendering` is the type that is produced by the `render` method: it
+    /// is commonly a view / screen model.
+    associatedtype Rendering
+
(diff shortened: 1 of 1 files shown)


Workflow/Sources/Workflow.swift
```diff
@@ -49,37 +49,71 @@
 /// }
 /// ```
 ///
-public protocol Workflow: AnyWorkflowConvertible {
-    /// Defines the state that is managed by this workflow.
-    associatedtype State
+#if swift(>=5.7)
+    public protocol Workflow<Rendering, Output>: AnyWorkflowConvertible {
```
jamieQ, line 53: for the time being, decided to just make this change on `Workflow` itself, as i believe that is by far the most useful change. if there are other compelling additions we could add them in the future.

with this change, we can now leverage the following forms of type erasure:

```swift
func boxedAnyWorkflow() -> AnyWorkflow<Rendering, Output> {
  MyWorkflow().asAnyWorkflow()
}

func existentialAnyWorkflow() -> any Workflow<Rendering, Output> {
  if someCondition {
    return MyWorkflow()
  } else {
    return MyOtherWorkflow()
  }
}

func opaqueWorkflow() -> some Workflow<Rendering, Output> {
  MyWorkflow()
}
```

mjohnson12 (approved):
Nice addition.

PR #185, opened 2023-01-31
Abstract ViewEnvironment to shared framework

n8chur:
## Overview

Environment value objects and corresponding propagation systems have become a useful tool in application architecture. Implementations of this tool can be found in [SwiftUI](https://developer.apple.com/documentation/swiftui/environmentvalues), [Blueprint](https://github.com/square/Blueprint/blob/89c0d793de404ac35b31026fdb64744c0a782ddb/BlueprintUI/Sources/Environment/Environment.swift), [WorkflowUI](https://github.com/square/workflow-swift/blob/0f7aca54f3d218c47cf974eb6d22f737b65fe204/WorkflowUI/Sources/Screen/ViewEnvironment/ViewEnvironment.swift), and the Market Design System's  [MarketEnvironment](https://github.com/squareup/market/tree/c8ed57955b3bbc44e333dff57ce932f7a47fecde/ios/MarketEnvironment)/[MarketEnvironmentUI](https://github.com/squareup/market/tree/c8ed57955b3bbc44e333dff57ce932f7a47fecde/ios/MarketEnvironmentUI).

In building out Market, we've found that there is great utility in consolidating these environments, particularly when their conceptual domain for propagation directly overlaps. Workflow's `ViewEnvironment` and Market's `MarketEnvironment` is a great example of this—both define an environment value bucket which is propagated through `UIViewController` hierarchies.

Needing to manage two types that attempt to solve an extremely similar goal leads to a few ares of complexity:

### Bridging

Bridging between the two propagation systems is more tedious than it need be since we're forced to acknowledge all properties we would like to bridge from one environment to the other at their boundaries (e.g. a vanilla UIViewController hierarchy that uses an environment propagation system which contains a child `WorkflowHostingController` that needs that environment information for theming).

It'd be much more convenient if we could just pass the entire environment along and avoid the need to build protocol abstractions over their shared API.

We could even eventually provide official support for automatically bridging between UIViewController and Workflow propagation of the environment within `WorkflowUI` (e.g. in `WorkflowHostingController` or a similar type), which would reduce a huge pain point and common mistake found in the `ios-register` codebase.

### Generics

We use protocols to abstract similar concepts which are shared amongst our environments (e.g. the presence of a `theme` property, the ability to resolve a modal presentation style from an Environment, etc.). This leads to the use of generics upstream of these APIs, which adds unnecessary clutter when building features which intersect with these concerns.

There is a particular use-case with our Modals framework which would be drastically simplified and cleaned up if we could share an environment type for all use cases which a Modal presentation style could be resolved.

### Code duplication

We currently duplicate all of our `ViewEnvironment` key conformances and extensions for use in `MarketEnvironment`. It'd be nice not to have to maintain both.

## Why is abstracting `ViewEnvironment` to a separate framework important? Why not just `import WorkflowUI` and use it where this reuse would be convenient?

While `Workflow`/`WorkflowUI` is a fairly lightweight framework, we've had requests not to depend on all of `Workflow`/`WorkflowUI` in some contexts. We'd prefer the reliance on a dependency like `Workflow`/`WorkflowUI` not to be the blocking factor in adopting the Market design system. This abstracted `ViewEnvironment` framework will completely replace our existing `MarketEnvironment` framework, so binary size etc. should not be increased as a result of this change in those use cases.

## Should we consider pulling `ViewEnvironment` out into a separate repository?

Probably! Just pulling this dependency out in Workflow was the easiest way to achieve our immediate needs for simplification so we decided to start here, but there may be utility in sharing `ViewEnvironment` in other contexts.

## Checklist

- [x] Unit Tests
- [x] UI Tests


Development.podspec
```diff
@@ -15,6 +15,7 @@ Pod::Spec.new do |s|
   s.dependency 'WorkflowRxSwift'
   # s.dependency 'WorkflowCombine' # TODO: Disabled because app specs cannot increase the deployment target of the root
```
kyleve, line 16: Unrelated, but I bet we can turn these on globally now cc @jamieQ

ViewEnvironment.podspec
```diff
@@ -0,0 +1,22 @@
+require_relative('version')
+
+Pod::Spec.new do |s|
```
kyleve, line 3: Any tests to move too?

kyleve (approved):

Development.podspec
n8chur: I don't believe we can if this is to work around a bug in `pod gen` unless we're bumping all podspecs to iOS 14.

The project attempts to generate with an iOS 11 deployment target and fails to resolve all targets as a result (since these commented ones require iOS 14).  I just reported the bug in #cocoapods [here](https://square.slack.com/archives/C09R3231Q/p1675121889596839).

kyleve: I figured we could bump stuff in the whole repo to 14 as well. Anyway, not important for this PR!

n8chur: I'd expect that you'd be able generate a project with pod gen which ends up just using the _highest_ deployment target found in your development podspecs, but it doesn't appear to work this way (it's either an order thing or choosing the lowest version I think).

ViewEnvironment.podspec
n8chur: I didn't see any but I thought about adding a simple `.empty` initialization test.

n8chur:
I didn't realize a draft would tag folks for review—I'll be updating this PR with a description of what and why this change is desirable and open it for full review soon!

The TL;DR Is that we'd get some pretty huge benefit from being able to share the `ViewEnvironment` type in our vanilla, Workflow-less `UIViewController` propagation system we use to propagate theming information outside of Workflow in Market.

WorkflowUI/Sources/ModuleExports.swift
```diff
@@ -0,0 +1 @@
+@_exported import ViewEnvironment
```
nononoah, line 1: This is a compatibility gesture, right? Assuming that `Foo` refers to the same module, can two modules call export `Foo` without conflict?

My hunch is that it's fine, and that there's probably a lot of `@_exported Foundation` lurking around, for example, but I figured it worth asking.

nononoah (approved):

WorkflowUI/Sources/ModuleExports.swift
n8chur: > This is a compatibility gesture, right?

Yup!

> can two modules call export `Foo` without conflict?

I believe so

ViewEnvironment.podspec
```diff
@@ -0,0 +1,22 @@
...
+    s.summary      = 'A collection of environment values propagated through a view hierarchy.'
+    s.homepage     = 'https://www.github.com/square/workflow-swift'
+    s.license      = 'Apache License, Version 2.0'
+    s.author       = 'Square'
+    s.source       = { :git => 'https://github.com/square/workflow-swift.git', :tag => "v#{s.version}" }
+
+    # 1.7 is needed for `swift_versions` support
+    s.cocoapods_version = '>= 1.7.0'
+
+    s.swift_versions = ['5.0']
```
lickel, line 15: Nit: move these things into `version.rb` ?

RELEASING.md
```diff
@@ -47,6 +47,7 @@ For Squares, membership is managed through the `Workflow Swift Owners` registry
     bundle exec pod trunk push WorkflowSwiftUI.podspec --synchronous
     bundle exec pod trunk push WorkflowCombine.podspec --synchronous
     bundle exec pod trunk push WorkflowCombineTesting.podspec --synchronous
+    bundle exec pod trunk push ViewEnvironment.podspec --synchronous
```
watt, line 50: These are ordered by dependency — you'll have to push this one first (or at least before `WorkflowUI`).

watt (approved):

Development.podspec
jamieQ: looking into bumping the versions [here](https://github.com/square/workflow-swift/pull/186). can do that after this is merged though & will update the values in the new podspec.

jamieQ (approved):

ViewEnvironment.podspec
n8chur: Seems reasonable to me—but I'm going to leave it as for this PR to avoid conflating that change here.

RELEASING.md
n8chur: Good catch, [done](https://github.com/square/workflow-swift/pull/185/commits/f69e7a790ef09813a77d4fbbc5dc716f8d8ab885)!

PR #188, opened 2023-02-03
[feat]: add primary associated types to more protocols

jamieQ:
### Description

since we've bumped to swift 5.7, let's add primary associated types in some more places

- adds `<Rendering, Output>` PAT to `AnyWorkflowConvertible`
- adds `<WorkflowType>` PAT to `WorkflowAction`
- adds `<Output>` to the various definitions of `Worker`

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowCombine/Sources/Worker.swift
+++ b/WorkflowCombine/Sources/Worker.swift
@@ -28,7 +28,7 @@ import Workflow
 /// If there is, and if the workers are 'equivalent', the context leaves the existing worker running.
 ///
 /// If there is not an existing worker of this type, the context will kick off the new worker (via `run`).
-public protocol Worker: AnyWorkflowConvertible where Rendering == Void {
+public protocol Worker<Output>: AnyWorkflowConvertible where Rendering == Void {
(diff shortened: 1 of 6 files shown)


kyleve (approved):

WorkflowCombine/Sources/Worker.swift
```diff
@@ -28,7 +28,7 @@ import Workflow
 /// If there is, and if the workers are 'equivalent', the context leaves the existing worker running.
 ///
 /// If there is not an existing worker of this type, the context will kick off the new worker (via `run`).
-public protocol Worker: AnyWorkflowConvertible where Rendering == Void {
+public protocol Worker<Output>: AnyWorkflowConvertible where Rendering == Void {
     /// The type of output events returned by this worker.
     associatedtype Output
     associatedtype WorkerPublisher: Publisher where
```
jamieQ, line 34: i assume we don't want this one to also be a primary associated type, but do correct me if there's some use case in which that would be valuable that i'm overlooking

mjohnson12 (approved):

PR #189, opened 2023-02-06
[fix]: use weak reference to internal sinks when vending to clients

jamieQ:
### Description

previously we would allow a strong reference to the internal `ReusableSink` to be captured by the `Sink` types that are vended to clients. this meant that events sent into such 'external' sinks would still propagate through the event handling machinery a bit before realizing that the backing workflow node to which they were destined was no longer around. by using weak references we can prevent extending the life of the internal sinks (and underlying EventPipe) unnecessarily in such cases.

- make `ReusableSink` captures `weak` when vending to the 'outside world'
- update various access control values to facilitate testing this behavior

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -198,12 +198,11 @@ extension WorkflowNode.SubtreeManager {
         func makeSink<Action>(of actionType: Action.Type) -> Sink<Action> where Action: WorkflowAction, WorkflowType == Action.WorkflowType {
             let reusableSink = sinkStore.findOrCreate(actionType: Action.self)
 
-            let signpostRef = SignpostRef()
+            let sink = Sink<Action> { [weak reusableSink] action in
+                WorkflowLogger.logSinkEvent(ref: SignpostRef(), action: action)
 
-            let sink = Sink<Action> { action in
-                WorkflowLogger.logSinkEvent(ref: signpostRef, action: action)
-
-                reusableSink.handle(action: action)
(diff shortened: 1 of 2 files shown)


Workflow/Sources/SubtreeManager.swift
```diff
@@ -198,12 +198,11 @@ extension WorkflowNode.SubtreeManager {
         func makeSink<Action>(of actionType: Action.Type) -> Sink<Action> where Action: WorkflowAction, WorkflowType == Action.WorkflowType {
             let reusableSink = sinkStore.findOrCreate(actionType: Action.self)
 
-            let signpostRef = SignpostRef()
```
jamieQ, line 201: unrelated change, but figured we might as well not instantiate this here since it isn't necessarily going to be used (we could also further improve this with some additional changes).

Workflow/Tests/SubtreeManagerTests.swift
```diff
@@ -243,6 +243,26 @@ final class SubtreeManagerTests: XCTestCase {
         XCTAssertEqual(manager.sideEffectLifetimes.count, 1)
         XCTAssertEqual(manager.sideEffectLifetimes.keys.first, "key-2")
     }
+
+    func test_eventPipes_notRetainedByExternalSinks() {
```
jamieQ, line 247: confirmed this fails prior to the change introduced in this PR

JaviSoto (approved):

PR #190, opened 2023-02-07
[chore]: refactor some internal actions to use existential any

jamieQ:
### Description

since we now have access to primary associated types, there are a couple places internally where it likely makes sense to leverage the language's 'existential' boxing facilities over our own. to start, we'll update a couple places that pass an `AnyWorkflowAction<T>` that can now be replaced with `any WorkflowAction<T>`.

additional comments inline.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -114,7 +114,7 @@ extension WorkflowNode {
 
 extension WorkflowNode.SubtreeManager {
     enum Output {
-        case update(AnyWorkflowAction<WorkflowType>, source: WorkflowUpdateDebugInfo.Source)
+        case update(any WorkflowAction<WorkflowType>, source: WorkflowUpdateDebugInfo.Source)
         case childDidUpdate(WorkflowUpdateDebugInfo)
     }
 }
@@ -175,7 +175,7 @@ extension WorkflowNode.SubtreeManager {
                 /// Update the existing child
                 existing.update(
                     workflow: workflow,
-                    outputMap: { AnyWorkflowAction(outputMap($0)) },
+                    outputMap: { outputMap($0) },
                     eventPipe: eventPipe
                 )
                 child = existing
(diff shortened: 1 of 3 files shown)


Workflow/Sources/WorkflowNode.swift
```diff
@@ -48,9 +48,19 @@ final class WorkflowNode<WorkflowType: Workflow> {
...
 
         switch subtreeOutput {
-        case .update(let event, let source):
-            /// Apply the update to the current state
-            let outputEvent = event.apply(toState: &state)
+        case .update(let action, let source):
+
+            /// 'Opens' the existential `any WorkflowAction<WorkflowType>` value
+            /// allowing the underlying conformance to be applied to the Workflow's State
+            func openAndApply<A: WorkflowAction>(
```
jamieQ, line 55: it's not entirely clear to me why this is necessary, or exactly why it works, but if you try to invoke `action.apply(...)` on the `any WorkflowAction` variable, you get a compiler error. this makes some amount of sense b/c an `any WorkflowAction` isn't itself a `WorkflowAction`. i did skim [this](https://github.com/apple/swift-evolution/blob/main/proposals/0352-implicit-open-existentials.md) and [this](https://forums.swift.org/t/using-openexistential-to-construct-a-rawrepresentable-type/48288/4) to see if there are other options, but this was the only thing i've been able to get working thus far.

Workflow/Sources/SubtreeManager.swift
```diff
@@ -272,7 +272,7 @@ extension WorkflowNode.SubtreeManager {
 
     fileprivate final class ReusableSink<Action: WorkflowAction>: AnyReusableSink where Action.WorkflowType == WorkflowType {
         func handle(action: Action) {
-            let output = Output.update(AnyWorkflowAction(action), source: .external)
+            let output = Output.update(action, source: .external)
```
jamieQ, line 275: this change will help the [observation API](https://github.com/square/workflow-swift/pull/168) be more useful as we will be able to expose the underlying conformance and not an AnyWorkflow wrapper

mjohnson12 (approved):

PR #192, opened 2023-02-22
Improve introspection for SignalProducerWorkflow actions

amorde:
This change improves the usability of https://github.com/square/workflow-swift/pull/168 and also makes it easier to debug when inspecting arbitrary workflow actions.

Below is an example of the string interpolation result of a `SignalProducerWorkflow` action before and after this change:

```swift
func workflowDidReceiveAction<Action: WorkflowAction>(
    _ action: Action,
    workflow: Action.WorkflowType,
    session: WorkflowSession
) {
    print("\(action)")
}

// before
"AnyWorkflowAction<SignalProducerWorkflow<Output>>(_apply: (Function))"
// after
"Action(output: OrdersSDKImpl.UpcomingOrdersMutationWorker.Output.ordersDidMutate)"
```

The primary reason the output was not useful before was the usage of `AnyWorkflowaction(sendingOutput:)` which wraps the output value in a closure so it becomes inaccessible. https://github.com/square/workflow-swift/blob/8b7e97b79fdf1a28b9df97e785047762ff8ee93f/Workflow/Sources/WorkflowAction.swift#L67-L71

## Checklist

- [ ] Unit Tests
- [ ] UI Tests


kyleve (approved):

mjohnson12 (approved):

WorkflowReactiveSwift/Sources/SignalProducerWorkflow.swift
```diff
@@ -47,18 +47,27 @@ struct SignalProducerWorkflow<Value>: Workflow {
     public typealias State = Void
     public typealias Rendering = Void
 
+    struct Action: WorkflowAction {
```
jamieQ, line 50: this seems like a good change to improve the utility of type-level introspection. do you think repeating the associated workflow's type in the action's type itself would be valuable? i.e. `SignalProducerWorkflowAction` or something. or putting that info into the string via 'custom string convertible' or whatever the appropriate conformance may be.

jamieQ (approved):

WorkflowReactiveSwift/Sources/SignalProducerWorkflow.swift
amorde: I did consider that yeah. I'll go ahead and make that change.

I also considered pulling a generic `SendOutputAction` out as a standalone type, but wasn't sure if that was worth it

PR #193, opened 2023-02-23
Expose AnyScreen.wrappedScreen for inspection

amorde:
When using observation APIs from https://github.com/square/workflow-swift/pull/168 encountering an instance of `AnyScreen` leaves you with no options for retrieving the underlying type and doing something with it.

```swift
func workflowWillRender<WorkflowType: Workflow>(
    _ workflow: WorkflowType,
    state: WorkflowType.State,
    session: WorkflowSession
) -> ((WorkflowType.Rendering) -> Void)? {
    return { (rendering: WorkflowType.Rendering) -> Void in
        guard let screen = rendering as? Screen else {
            return
        }

        if screen is AnyScreen {
            // ?? Can't do anything here since we cannot access `.wrappedScreen`
        }
    }
}        
```

This can also come up in `ScreenViewController.update(screen:environment:)` when using Screen types that wrap other screen values, such as [BackStackScreen](https://github.com/square/workflow-swift/blob/dbdcb967b4c83cdc283b4bfd672c60f4398bd80b/Samples/BackStackContainer/Sources/BackStackScreen.swift#L19) if typed as `BackStackScreen<AnyScreen>`.

 This exposes the wrapped value so it can be accessed and casted as desired.

```swift

struct MyScreen: Screen {
    let objectCount: Int
}

func workflowWillRender<WorkflowType: Workflow>(
    _ workflow: WorkflowType,
    state: WorkflowType.State,
    session: WorkflowSession
) -> ((WorkflowType.Rendering) -> Void)? {
    return { (rendering: WorkflowType.Rendering) -> Void in
        guard let screen = rendering as? Screen else {
            return
        }

        // First wrap in AnyScreen if not one already. It won't get nested
        switch screen.asAnyScreen().wrappedScreen {
        case let typed as MyScreen:
            print("Workflow \(workflow) produced screen \(MyScreen.self) with object count \(typed.objectCount)")
            if typed.objectCount > 100 {
                Logger.instance.logError("Produced rendering with oversized object count: \(typed.objectCount)")
            }
        case let typed as SomeOtherScreen:
            print("Workflow \(workflow) produced unexpected screen \(typed)")
        }
    }
}
```

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
+++ b/WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
@@ -20,10 +20,7 @@ import UIKit
 
 public struct AnyScreen: Screen {
     /// The original screen, retained for debugging
-    internal let wrappedScreen: Screen
-
-    /// Stored getter for the wrapped screen’s view controller description
-    private let _viewControllerDescription: (ViewEnvironment) -> ViewControllerDescription
+    public let wrappedScreen: Screen
 
     public init<T: Screen>(_ screen: T) {
         if let anyScreen = screen as? AnyScreen {
@@ -31,12 +28,10 @@ public struct AnyScreen: Screen {
             return
         }
         self.wrappedScreen = screen
-        self._viewControllerDescription = screen.viewControllerDescription(environment:)
     }
 
     public func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
-        // Passed straight through
-        return _viewControllerDescription(environment)
+        return wrappedScreen.viewControllerDescription(environment: environment)
     }
 }
 


WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
```diff
@@ -20,23 +20,18 @@ import UIKit
...
             return
         }
         self.wrappedScreen = screen
-        self._viewControllerDescription = screen.viewControllerDescription(environment:)
     }
 
     public func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
-        // Passed straight through
-        return _viewControllerDescription(environment)
+        return wrappedScreen.viewControllerDescription(environment: environment)
```
amorde, line 34: this is an unrelated cleanup. I couldn't see a reason why this abstraction was necessary. It looks like it's been there since the type was [introduced](https://github.com/square/workflow-swift/commit/fc93feb64e3ad7cdf66e28dc6ee17b566662024d) but at the time the view registry existed so perhaps that had something to do with it

kyleve (approved):
This seems reasonable; but I also wonder if we want to solve this in a way that doesn't require checking for any screen explicitly; Eg should screen have some sort of "loggingInfo" property on it that defaults to just returning the screen type, and AnyScreen could override this to return wrappedScreen.loggingInfo?

amorde:
>  should screen have some sort of "loggingInfo" property on it that defaults to just returning the screen type

This might work for a specific case of trying to print out the screen type or something similar, but usually `Screen` conformers are structs with real values in them that you might want to access, and requiring every single `Screen` to add conformance to a new protocol to get the values out feels excessive (at least, given our internal usage of Workflow)

kyleve:
> > should screen have some sort of "loggingInfo" property on it that defaults to just returning the screen type
> 
> This might work for a specific case of trying to print out the screen type or something similar, but usually `Screen` conformers are structs with real values in them that you might want to access, and requiring every single `Screen` to add conformance to a new protocol to get the values out feels excessive (at least, given our internal usage of Workflow)

Sorry, I meant sticking that onto Screen itself with a default impl. Then AnyScreen could override that

amorde:
Right but what would `loggingInfo` return? Maybe I misunderstood but I'm imagining cases where you actually want to read values off of a screen in one of the observation APIs, so even if there was something like `loggingInfo: Screen.Type` that still wouldn't let me read values from the instance. 

I think I might be misunderstanding what you mean though.

amorde:
I've updated the PR example to include reading values form the screen to demonstrate what I mean

WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
jamieQ: hmm yeah also not sure, but seems fine to remove. if you weren't going to store the underlying screen, then i could see why it would exist, so maybe it was here first, and the stored screen value was added after?

WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
```diff
@@ -20,23 +20,18 @@ import UIKit
 
 public struct AnyScreen: Screen {
     /// The original screen, retained for debugging
-    internal let wrappedScreen: Screen
-
-    /// Stored getter for the wrapped screen’s view controller description
-    private let _viewControllerDescription: (ViewEnvironment) -> ViewControllerDescription
+    public let wrappedScreen: Screen
```
jamieQ, line 23: have a slight inclination to put this behind an `_spi` annotation, since it seems like peering into this value should be kind of an exception to typical use.

WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
amorde: I thought the same thing but the original implementation stored the screen
```
public struct AnyScreen: Screen {
    internal let wrappedScreen: Screen
    private let viewControllerBuilder: (ViewRegistry) -> AnyScreenViewController.WrappedViewController

    public init<T: Screen>(_ screen: T) {
        self.wrappedScreen = screen
        self.viewControllerBuilder = { viewRegistry in
            return viewRegistry.provideView(for: screen)
        }
    }

    // Internal method to inflate the wrapped screen from a view registry
    func makeViewController(from viewRegistry: ViewRegistry) -> AnyScreenViewController.WrappedViewController {
        return viewControllerBuilder(viewRegistry)
    }
}
```

jamieQ (approved):
this also seems reasonable to me. have a mild preference to grant access to the value via SPI.

WorkflowUI/Sources/Screen/AnyScreen/AnyScreen.swift
amorde: it feels like a standard thing for type erasers to expose the original value. for example Swift's AnyHashable exposes the original value https://developer.apple.com/documentation/swift/anyhashable/base

kyleve: I was gonna say; should we rename this to `base` for consistency?

PR #194, opened 2023-03-14
[task]: attempt to improve debugging info for AnyWorkflowAction

jamieQ:
### Description

this change adds introspection capabilities for what i believe is the final type-eraser that lacks them – AnyWorkflowAction. changes include:

- adds a public `base` property on `AnyWorkflowAction` to expose the underlying action as an `Any` value
- adds an internal utility type, `ClosureAction`, for closure-based `AnyWorkflowAction`s. the intent is to provide the possibility of improved telemetry information when logging such actions.
- adds a new public property `isClosureBased` on `AnyWorkflowAction` to differentiate instances that forward to an underlying wrapped action vs those whose implementations are defined by a closure.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Workflow/Sources/WorkflowAction.swift
+++ b/Workflow/Sources/WorkflowAction.swift
@@ -39,6 +39,9 @@ public struct AnyWorkflowAction<WorkflowType: Workflow>: WorkflowAction {
     /// The underlying type-erased `WorkflowAction`
     public let base: Any
 
+    /// True iff the underlying `apply` implementation is defined by a closure vs wrapping a `WorkflowAction` conformance
+    public let isClosureBased: Bool
+
     /// Creates a type-erased workflow action that wraps the given instance.
     ///
     /// - Parameter base: A workflow action to wrap.
@@ -49,6 +52,7 @@ public struct AnyWorkflowAction<WorkflowType: Workflow>: WorkflowAction {
         }
         self._apply = { return base.apply(toState: &$0) }
         self.base = base
+        self.isClosureBased = false
     }
 
     /// Creates a type-erased workflow action with the given `apply` implementation.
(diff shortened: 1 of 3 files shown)


Workflow/Sources/WorkflowAction.swift
```diff
@@ -46,13 +48,30 @@ public struct AnyWorkflowAction<WorkflowType: Workflow>: WorkflowAction {
...
     ///
     /// - Parameter apply: the apply function for the resulting action.
-    public init(_ apply: @escaping (inout WorkflowType.State) -> WorkflowType.Output?) {
-        self._apply = apply
+    public init(
+        _ apply: @escaping (inout WorkflowType.State) -> WorkflowType.Output?,
+        fileID: StaticString = #fileID,
+        line: UInt = #line
+    ) {
+        let closureAction = ClosureAction<WorkflowType>(
```
amorde, line 62: could we expose a way for consumers who receive `AnyWorkflowAction` to know if this is wrapping a closure vs. a real action? could maybe make `ClosureAction` public to achieve that

```
let action: AnyWorkflowAction<SomeWorkflow> = ...

if action.base is ClosureAction<SomeWorkflow> {

}
```

jamieQ: yeah we can do that. do you think full access to the underlying type is needed? or would a flag suffice?

mjohnson12 (approved):

Workflow/Sources/WorkflowAction.swift
amorde: In general my default is "provide more access" but I know as the SDK author its the opposite for you :p for now I suppose a flag would work

amorde (approved):

Workflow/Sources/WorkflowAction.swift
```diff
@@ -78,3 +102,34 @@ extension AnyWorkflowAction {
...
 }
+
+// MARK: Closure Action
+
+/// A `WorkflowAction` that wraps an `apply(...)` implementation defined by a closure.
+/// Mainly used to provide more useful debugging/telemetry information for `AnyWorkflow` instances
+/// defined via a closure.
+struct ClosureAction<WorkflowType: Workflow>: WorkflowAction {
+    private let _apply: (inout WorkflowType.State) -> WorkflowType.Output?
+    let fileID: StaticString
```
amorde, line 113: nice touch!

PR #106, opened 2021-11-12
Add EnvironmentScreen and associated Screen modifiers

bencochran:
Is it time? (See https://github.com/square/workflow/pull/999#discussion_r387871586 & https://github.com/square/workflow/issues/1001)

## Checklist

- [X] Unit Tests
- [X] I have made corresponding changes to the documentation

PR #195, opened 2023-03-14
Introduce AdaptedEnvironmentScreen, for modifying the ViewEnvironment when creating screens

kyleve:
This mirrors the API allowed on `Elements` in blueprint, to ease `ViewEnvironment` modification. For example, you might do something like this:

```swift
func render(...) -> Rendering {
    MarketSelectSheetScreen( ... construct your select actions ...)
        .adaptedEnvironment(keyPath: \.isContentAreaEnabled, value: state.isSaving == false)
}
```

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
+++ b/WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2023 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+
+import Foundation
+import ViewEnvironment
+
+/// Wraps a `Screen` tree with a modified `ViewEnvironment`.
+///
+/// By specifying environmental values with this `Screen`, all child screens nested
+/// will inherit those values automatically. Values can be changed
+/// anywhere in a sub-tree by inserting another `AdaptedEnvironmentScreen`.
+///
+/// ```swift
+/// MyScreen(...)
+///     .adaptedEnvironment(keyPath: \.myValue, to: newValue)
+/// ```
+///
+public struct AdaptedEnvironmentScreen<Content: Screen>: Screen {
+    /// The screen wrapped by this screen.
+    public var wrapped: Content
+
+    /// Takes in a mutable `ViewEnvironment` which can be mutated to add or override values.
+    public typealias Adapter = (inout ViewEnvironment) -> Void
+
+    var adapter: Adapter
+
+    /// Wraps a `Screen` with an environment that is modified using the given configuration block.
+    ///
+    /// - Parameters:
+    ///   - wrapping: The screen to be wrapped.
+    ///   - adapting: A block that will set environmental values.
+    public init(
+        wrapping wrapped: Content,
+        adapting: @escaping Adapter
+    ) {
+        self.wrapped = wrapped
+        self.adapter = adapting
+    }
+
+    /// Wraps a `Screen` with an environment that is modified for a single key and value.
+    ///
+    /// - Parameters:
+    ///   - wrapping: The screen to be wrapped.
+    ///   - key: The environment key to modify.
+    ///   - value: The new environment value to cascade.
+    public init<Key: ViewEnvironmentKey>(
+        wrapping screen: Content,
+        key: Key.Type,
+        value: Key.Value
+    ) {
+        self.init(wrapping: screen, adapting: { $0[key] = value })
+    }
+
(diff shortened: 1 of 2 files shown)


WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
```diff
@@ -0,0 +1,110 @@
+//
+//  File.swift
```
square-tomb, line 2: A real file's file

WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
```diff
@@ -0,0 +1,110 @@
...
+/// ```
+///
+public struct AdaptedEnvironmentScreen<Content: Screen>: Screen {
+    /// The screen wrapped by this screen.
+    public var wrapped: Content
+
+    /// Takes in a mutable `ViewEnvironment` which can be mutated to add or override values.
+    public typealias Adapter = (inout ViewEnvironment) -> Void
+
+    var adapter: Adapter
```
square-tomb, line 31: Doesn't really matter because this is a value type, but why can't these be `let`? It doesn't look we're using their `var`ness in the implementation, tests, or sample usage.

square-tomb (approved):

WorkflowUI/Tests/AdaptedEnvironmentScreenTests.swift
```diff
@@ -0,0 +1,53 @@
...
+        let screen = TestScreen { environment = $0 }
+            .adaptedEnvironment(key: TestingKey1.self, value: "adapted1.1")
+            .adaptedEnvironment(key: TestingKey1.self, value: "adapted1.2")
+            .adaptedEnvironment(key: TestingKey2.self, value: "adapted2.1")
+            .adaptedEnvironment(key: TestingKey1.self, value: "adapted1.3")
+            .adaptedEnvironment(key: TestingKey2.self, value: "adapted2.2")
+
+        _ = screen.viewControllerDescription(environment: .empty)
+
+        // The inner-most change; the one closest to the screen; should be the value we get.
```
jamieQ, line 27: at first glance, this seems a bit confusing, and almost feels backwards. i can see why it is true and expected (the adapters are applied from 'outside in'), but it makes the example code above seem more confusing. since the outer wrappers that attempt to adapt earlier keys effectively do nothing, does it make sense to allow them to be written? should there be some form of warning produced if clients attempt to do this? as implemented here, this might not be possible to detect, but just something maybe worth considering.

WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
kyleve: Ugh, I hate how SPM just makes a file called "File". Removing this and adding the boilerplate.

WorkflowUI/Tests/AdaptedEnvironmentScreenTests.swift
```diff
@@ -0,0 +1,53 @@
...
+
+import UIKit
+import XCTest
+@testable import WorkflowUI
+
+class AdaptedEnvironmentScreenTests: XCTestCase {
+    func test_wrapping() {
+        var environment: ViewEnvironment = .empty
+
+        let screen = TestScreen { environment = $0 }
```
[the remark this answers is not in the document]
kyleve: It will be, yeah. I think that's generally fine; since deeply nesting like this in practice won't really occur. This test is a bit of a degenerate test, designed to simulate many `adaptedEnvironments` that could be throughout your Screen tree.

> there are a few alternatives i can think of for accumulating the environment mutations

We do this in Blueprint, but its a _lot_ easier because the `AdaptedEnvironment` element there isn't generic around the element type; the generics make this much more difficult.

WorkflowUI/Tests/AdaptedEnvironmentScreenTests.swift
kyleve: I mentioned this above; but this is basically a somewhat artificial test / a test designed to simulate where these adaptations happen throughout the screen tree; to make sure the "last write wins" is true.

> should there be some form of warning produced if clients attempt to do this

I don't think this is really possible; without _also_ including the `ViewEnvironmentKey` in the generic of the screen; which then also means we'd need to get rid of the keypath and closure-based modifiers.

jamieQ (approved):

WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
kyleve: I tend to just keep things as `var` unless `let` is absolutely required (eg, consistency between variables), mostly as a purely stylistic thing.

WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
```diff
@@ -0,0 +1,119 @@
...
+/// By specifying environmental values with this `Screen`, all child screens nested
+/// will inherit those values automatically. Values can be changed
+/// anywhere in a sub-tree by inserting another `AdaptedEnvironmentScreen`.
+///
+/// ```swift
+/// MyScreen(...)
+///     .adaptedEnvironment(keyPath: \.myValue, to: newValue)
+/// ```
+///
+public struct AdaptedEnvironmentScreen<Content: Screen>: Screen {
```
kyleve, line 33: Looking at https://github.com/square/workflow-swift/pull/106 from @bencochran, I think I should likely move this conformance to an extension.

kyleve: https://github.com/square/workflow-swift/pull/198

PR #198, opened 2023-03-23
Conditionally conform AdaptedEnvironmentScreen to Screen, based on its content

kyleve:
Based on the work done in https://github.com/square/workflow-swift/pull/106, so that way you can do an adaptation before, eg you map something to a `Screen`.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
+++ b/WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
@@ -30,7 +30,7 @@ import ViewEnvironment
 ///     .adaptedEnvironment(keyPath: \.myValue, to: newValue)
 /// ```
 ///
-public struct AdaptedEnvironmentScreen<Content: Screen>: Screen {
+public struct AdaptedEnvironmentScreen<Content> {
     /// The screen wrapped by this screen.
     public var wrapped: Content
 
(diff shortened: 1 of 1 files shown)


WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
```diff
@@ -30,7 +30,7 @@ import ViewEnvironment
 ///     .adaptedEnvironment(keyPath: \.myValue, to: newValue)
 /// ```
 ///
-public struct AdaptedEnvironmentScreen<Content: Screen>: Screen {
+public struct AdaptedEnvironmentScreen<Content> {
     /// The screen wrapped by this screen.
     public var wrapped: Content
```
jamieQ, line 35: nit: the parameter names & comments now seem slightly mis-matched if the generic type is unconstrained

jamieQ (approved):

WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
```diff
@@ -30,7 +30,7 @@ import ViewEnvironment
 ///     .adaptedEnvironment(keyPath: \.myValue, to: newValue)
 /// ```
 ///
-public struct AdaptedEnvironmentScreen<Content: Screen>: Screen {
```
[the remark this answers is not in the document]
kyleve: I'll defer to @bencochran here since he's the one who usually structures things this way, but eg, it's common to build a back stack wrapping some type of view model, etc, and then map that into a screen type after you fetch it. This allows you to use environment adaptation alongside that.

WorkflowUI/Sources/Screen/AdaptedEnvironmentScreen.swift
kyleve: I think in reality, this is going to be 99% screens / mapped to one eventually, so I left it as is.

PR #205, opened 2023-05-01
[feat]: add ViewEnvironmentUI module

n8chur:
> **Note**: This PR is targeting the [`feature/viewenvironmentui`](https://github.com/square/workflow-swift/tree/feature/viewenvironmentui) branch. Further integrations into `WorkflowUI` will land in that branch before these changes are merged into `main` (#211).

# Overview

Adds a new `ViewEnvironmentUI` module which handles propagation of `ViewEnvironment` through `UIViewController`, `UIView`, and any arbitrary object node.

# Changes

The changes in this PR are most easily reviewable by commit:

## Port ViewEnvironmentUI from Market and add to Development.podspec ([6eacf4a](https://github.com/square/workflow-swift/pull/205/commits/6eacf4ad78e3636cd456a6d414b9623100ea83ee))

This is essentially a copy paste of the framework which currently exists in MarketUI, but with a few typos in documentation fixed and added license header docs.

## Simplify protocol hierarchy ([7b7c59f](https://github.com/square/workflow-swift/pull/205/commits/7b7c59fd0543c01420c804327fbd337fd583d3f0))

Removes the `ViewEnvironmentCustomizing` protocol and merges the functions it defined into `ViewEnvironmentObserving`.

Adds default implementations of all functions on `ViewEnvironmentObserving` so you can pick and chose how you'd like to observe without requiring an empty implementation for functions you may not be concerned with.

Removes the ability to implement `var viewEnvironment: ViewEnvironment` yourself. This was useful when bridging between propagation contexts and/or when a node wanted to act as a root without pulling the environment through all ancestors above it. Instead, setting the nodes `environmentAncestor` to `nil` will prevent the pull from above and will prevent other nodes who have this node set as an ancestor from triggering `setNeedsEnvironmentUpdate` on the node.

## Rename viewEnvironment to environment ([593250d](https://github.com/square/workflow-swift/pull/205/commits/593250dca422716488ebb6c13b8223daad21a4b3))

This rename will cause ambiguity when using `ScreenViewController` if `ViewEnvironmentUI` is imported since that superclass already has an `environment` property. This will be resolved in a future PR targeting the feature branch.

## Add environmentDidChange observation option ([b37dad6](https://github.com/square/workflow-swift/pull/205/commits/b37dad6b1c913e4b578b1fd5cc05e2a14e812fbc))

Adds a hook that observers can utilize to respond aggressively to environment changes rather than needing to wait until the `setNeedsEnvironmentUpdate`/`apply(environment:)` dance has completed (which is usually deferred since it's backed by `setNeedsLayout`/`viewWillLayoutSubviews`).

# Checklist

- [x] Unit Tests
- [x] UI Tests


ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
```diff
@@ -0,0 +1,272 @@
...
+    /// `viewWillLayoutSubviews()` for `UIViewController`s and `layoutSubviews()` for `UIView`s.
+    ///
+    /// This will call ``apply(environment:)`` on the receiver if the node has been flagged for needing update.
+    ///
+    public func applyEnvironmentIfNeeded() {
+        guard needsEnvironmentUpdate else { return }
+
+        needsEnvironmentUpdate = false
+
+        if let observing = self as? ViewEnvironmentObserving {
```
kyleve, line 59: Do you need this `if let`, given the chain upwards of the protocol inheritance?

ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
```diff
@@ -0,0 +1,272 @@
...
+    /// Adds a `ViewEnvironment` change observation.
+    ///
+    /// The observation will only be active for as long as the returned lifetime is retained or
+    /// `cancel()` is called on it.
+    ///
+    @_spi(ViewEnvironmentWiring)
+    public func addEnvironmentNeedsUpdateObserver(
+        _ onNeedsUpdate: @escaping (ViewEnvironment) -> Void
+    ) -> ViewEnvironmentUpdateObservationLifetime {
+        let object = NSObject()
```
kyleve, line 76: Thought: Is it worth making a special type / subclass, so the memory debugger, etc, shows the type more obviously in case something goes wrong?

ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
```diff
@@ -0,0 +1,272 @@
...
+        get {
+            objc_getAssociatedObject(self, &AssociatedKeys.ancestorOverride) as? EnvironmentAncestorProvider
+        }
+        set {
+            assert(
+                newValue == nil
+                    || environmentAncestorOverride == nil,
+                "Attempting to set environment ancestor override when one is already set."
+            )
+            objc_setAssociatedObject(self, &AssociatedKeys.ancestorOverride, newValue, .OBJC_ASSOCIATION_RETAIN)
```
kyleve, line 137: Trying to think through if we can get into any weird cases by making this a retain, vs a weak.

kyleve (approved):
Couple questions; but approving since a lot of this has already existed in Market for the most part, and has been reviewed.

ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
n8chur: Yeah, this protocol does not inherit from `ViewEnvironmentObserving`.

n8chur: Nor does `ViewEnvironmentPropagatingObject`.

ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
n8chur: The type here is a closure so it'd be really easy to mess that up I think...

The fact that this property is retained is part of the trick I used in the prototype to add support for inserting a propagation node above the view controller being described in a `ViewControllerDescription`—this closure is the only thing holding a reference to that fake node which provides the bridged `ViewEnvironment`.

ViewEnvironmentUI/Sources/ViewEnvironmentObserving.swift
```diff
@@ -0,0 +1,92 @@
...
+    /// - Important: `UIViewController` and `UIView` conformers _must_ call
+    ///   ``ViewEnvironmentObserving/applyEnvironmentIfNeeded()-8gr5k``in `viewWillLayoutSubviews()` and
+    ///   `layoutSubviews()` respectively.
+    ///
+    func customize(environment: inout ViewEnvironment)
+
+    /// Consumers should apply the `ViewEnvironment` to their node when this function is called.
+    ///
+    /// - Important: `UIViewController` and `UIView` conformers _must_ call ``applyEnvironmentIfNeeded()-3bamq``
+    ///   in `viewWillLayoutSubviews()` and `layoutSubviews()` respectively.
```
watt, line 65: Do we need to have these callouts in `customize` and `apply`? I actually find it a little confusing to point it out here.

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -0,0 +1,144 @@
...
+    ///
+    /// This describes the ancestor that the `ViewEnvironment` is inherited from.
+    ///
+    /// To override the return value of this property for `UIViewController`/`UIView` subclasses, set the
+    /// ``ViewEnvironmentPropagatingObject/environmentAncestorOverride`` property.  If no override is present, the
+    /// return value will be `parent ?? presentingViewController`/`superview`.
+    ///
+    /// If the value of the ancestor is `nil`, by default, ancestors will not call notify this node of needing an
+    /// environment update as it changes. This allows a node to effectively act as a root node when needed (e.g.
+    /// bridging from other propagation systems like WorkflowUI).
```
watt, line 73:
```suggestion
    /// If the value of the ancestor is `nil`, by default, ancestors will not notify this node of needing an
    /// environment update as it changes. This allows a node to effectively detach from its ancestor and act as a root node when needed, without any changes to the ancestor (e.g.
    /// bridging from other propagation systems like WorkflowUI).
```

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -0,0 +1,144 @@
...
+        if let observing = self as? ViewEnvironmentObserving {
+            observing.customize(environment: &environment)
+        }
+
+        return environment
+    }
+
+    /// Notifies all appropriate descendants that the environment needs update.
+    ///
+    /// If a descendant's ancestor is `nil` it will not be notified of needing update.
```
watt, line 129:
```suggestion
    /// Ancestor-descendent bindings must be mutually agreed for this method to notify them. If a descendant's ancestor is `nil` it will not be notified of needing update.
```

ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
```diff
@@ -0,0 +1,272 @@
...
+}
+
+extension ViewEnvironmentPropagatingObject {
+    /// Adds a `ViewEnvironment` change observation.
+    ///
+    /// The observation will only be active for as long as the returned lifetime is retained or
+    /// `cancel()` is called on it.
+    ///
+    @_spi(ViewEnvironmentWiring)
+    public func addEnvironmentNeedsUpdateObserver(
```
watt, line 73: Do we still have a use case for this?

watt (approved):
Mostly tweaks to docs.

A couple thoughts on potential follow-ups:
- we should mob on diagrams for this some time
- might be nice to have a convenience function for inserting a customization node between existing ancestors & descendants

ViewEnvironmentUI/Sources/ViewEnvironmentPropagatingObject.swift
n8chur: I'm still using it in the prototype inside of `TrampolineModalListObserver`.

I think we may actually want to be using this method of observation instead of overriding `environmentDescendantsOverride` when the logic making this observation is not owned by the view controller itself.

For example, I'm currently playing with having `WorkflowHostingController` set its `environmentDescendantsOverride` to an empty array to prevent vanilla propagation invalidations causing a re-render of it's owned `rendering` since we'll already be performing a re-render on `environmentDidChange`. In our `TrampolineModalPresenter`, we need some way to observe invalidations triggered by its `owner`, but that owner could be a `WorkflowHostingController`. Rather than trying to override `environmentDescendantOverrides` and having to compose the existing override (similar to calling `super`), it might be cleaner to consider `environmentDescendantsOverride` something that only the VC it's applied to should override, and if you need to hook into invalidations, you use the `addEnvironmentNeedsUpdateObserver(...)`.

The composing approach does seem to work, but seems like it could get hard to follow when debugging and potentially difficult to document, but I'd be open to utilizing that approach for cases like this if we'd rather consolidate the API a bit.

ViewEnvironmentUI/Sources/ViewEnvironmentObserving.swift
n8chur: Nope, I'm fine removing them!

- [ ] Remove applyEnvironmentIfNeeded callouts in `customize` and `apply`

ViewEnvironmentUI/README.md
```diff
@@ -0,0 +1,6 @@
+# ViewEnvironmentUI
+
+`ViewEnvironmentUI` provides a means to propagate a `ViewEnvironment` through a heirarchy of object nodes.
```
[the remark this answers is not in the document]
n8chur: 117963c

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: 7fdbe94

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: 7fdbe94

n8chur:
> A couple thoughts on potential follow-ups:
> 
> * we should mob on diagrams for this some time
> * might be nice to have a convenience function for inserting a customization node between existing ancestors & descendants

Definitely down to pair on both of these at some point in the near future!

square-tomb (approved):
Thanks for the careful organization of this PR. That and the context in the linked proposal made it a pleasure to review.

My comments are all nits and doc fixes. I wonder if there is some way we can avoid the overloading of the term "ancestor" in our terminology.

ViewEnvironmentUI/Sources/ViewEnvironmentObserving.swift
```diff
@@ -49,17 +47,26 @@ import UIKit
 /// - Important: `UIViewController` and `UIView` conformers _must_ call ``applyEnvironmentIfNeeded()-3bamq``
 ///   in `viewWillLayoutSubviews()` and `layoutSubviews()` respectively.
 ///
-/// - Tag: ViewEnvironmentObserving
-///
-public protocol ViewEnvironmentObserving: ViewEnvironmentCustomizing {
+public protocol ViewEnvironmentObserving: ViewEnvironmentPropagating {
+    /// Customizes the `ViewEnvironment` as it flows through this propagation node to provide overrides to environment
+    /// values. These changes will be propagated to all descendant nodes.
```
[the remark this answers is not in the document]
n8chur: Customizations currently apply to the environment on the node itself. I'll clarify this in documentation.

- [ ] Add note about customizations applying to the node itself

Andrew mentioned that this was surprising during his review and I have [a note to follow up on this](https://www.notion.so/marketdesignsystem/Proposal-WorkflowUI-ViewEnvironment-Propagation-in-UIKit-0aec81d8c40545baa2ed7bacd938a6b7?pvs=4#4ae09b25017746a79959f67ec65b507e). Which shape is most intuitive to you?

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -16,6 +16,37 @@
 
 import ViewEnvironment
 
+/// `ViewEnvironmentPropagating` allows an environment propagation node to observe updates to the
+/// `ViewEnvironment` as it flows through the node hierarchy and have
+/// the environment applied to the node.
```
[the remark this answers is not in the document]
n8chur: Ah, yeah, this is describing `ViewEnvironmentObserving` behavior! Thanks for catching!

- [ ] Fix doc comment about `ViewEnvironmentPropagating` which described `ViewEnvironmentObserving` behavior

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -16,6 +16,37 @@
...
+///
+///     func apply(environment: ViewEnvironment) {
+///         // Apply values from the environment to your view controller (e.g. a theme)
+///     }
+///
+///     // If you'd like to override values in the environment you can provide them here. If you'd
+///     // like to just inherit the context from above there is no need to implement this function.
+///     func customize(environment: inout ViewEnvironment) {
+///         environment.traits.mode = .dark
+///     }
```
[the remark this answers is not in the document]
n8chur: https://github.com/square/workflow-swift/pull/205/files/6eacf4ad78e3636cd456a6d414b9623100ea83ee..7b7c59fd0543c01420c804327fbd337fd583d3f0#r1198033904

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -39,6 +70,10 @@ public protocol ViewEnvironmentPropagating {
     /// ``ViewEnvironmentPropagatingObject/environmentAncestorOverride`` property.  If no override is present, the
     /// return value will be `parent ?? presentingViewController`/`superview`.
     ///
+    /// If the value of the ancestor is `nil`, by default, ancestors will not call notify this node of needing an
+    /// environment update as it changes. This allows a node to effectively act as a root node when needed (e.g.
+    /// bridging from other propagation systems like WorkflowUI).
```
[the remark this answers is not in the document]
n8chur: Ah, yeah, we should probably be more careful with the phrasing here.

As you stated, "ancestor" is any parent (recursively), and "descendant" is children (recursively). A node's ancestor may not be that ancestor nodes' descendant and vice versa. I think this doc comment should be something like:

> If the value of the ancestor is `nil`, by default, other nodes configured with this node as a descendant will not notify this node of needing an environment update as it changes. This allows a node to effectively act as a root node when needed (e.g. bridging from other propagation systems like WorkflowUI).

Does that make sense?

- [ ] Improve documentation around the difference between ancestor and a node configured with the node in question as a descendant

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -0,0 +1,146 @@
...
+            observing.customize(environment: &environment)
+        }
+
+        return environment
+    }
+
+    /// Notifies all appropriate descendants that the environment needs update.
+    ///
+    /// Ancestor-descendent bindings must be mutually agreed for this method to notify them. If a descendant's ancestor
+    /// is `nil` it will not be notified of needing update.
```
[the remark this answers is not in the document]
n8chur: d918d9a

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -16,46 +16,19 @@
...
-///     // like to just inherit the context from above there is no need to implement this function.
-///     func customize(environment: inout ViewEnvironment) {
-///         environment.traits.mode = .dark
-///     }
-/// }
-/// ```
-///
-/// - Important: `UIViewController` and `UIView` conformers _must_ call ``applyEnvironmentIfNeeded()-3bamq``
-///   in `viewWillLayoutSubviews()` and `layoutSubviews()` respectively.
+/// Describes a node which supports `ViewEnvironment` propagation.
```
watt, line 19:
```suggestion
/// A node in a `ViewEnvironment` propagation tree.
///
/// This protocol describes the base functionality of every node in the tree:
/// - Reading the environment, via `environment`.
/// - Walking up the tree, via `environmentAncestor`.
/// - Walking down the tree, via `environmentDescendants`.
/// - Notifying a node that the environment changed, via `setNeedsEnvironmentUpdate`.
```

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -65,26 +38,27 @@ public protocol ViewEnvironmentPropagating {
...
-    /// ``ViewEnvironmentPropagatingObject/environmentAncestorOverride`` property.  If no override is present, the
-    /// return value will be `parent ?? presentingViewController`/`superview`.
+    /// `environmentAncestorOverride` property.  If no override is present, the return value will be `parent ?? 
+    /// `presentingViewController`/`superview`.
     ///
-    /// If the value of the ancestor is `nil`, by default, ancestors will not notify this node of needing an
-    /// environment update as it changes. This allows a node to effectively detach from its ancestor and act as a root
-    /// node when needed, without any changes to the ancestor (e.g. bridging from other propagation systems like
-    /// WorkflowUI).
+    /// If the value of the ancestor is nil, by default, other nodes configured with this node as a descendant will not
```
watt, line 44:
```suggestion
    /// Ancestor-descendent bindings must be mutually agreed. If the value of the ancestor is `nil`, then by default, other nodes configured with this node as a descendant will not
```

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -65,26 +38,27 @@ public protocol ViewEnvironmentPropagating {
...
     @_spi(ViewEnvironmentWiring)
     var environmentAncestor: ViewEnvironmentPropagating? { get }
 
-    /// The [`ViewEnvironment` propagation](x-source-tag://ViewEnvironmentObserving)
-    /// descendants.
+    /// The `ViewEnvironment` propagation descendants.
     ///
     /// This describes the descendants that will be notified when the `ViewEnvironment` changes.
+    /// 
+    /// If a descendant's `environmentAncestor` is `nil`, that descendant will not be notified when the
```
watt, line 55:
```suggestion
    /// Ancestor-descendent bindings must be mutually agreed. If a descendant's `environmentAncestor` is `nil`, that descendant will not be notified when the
```

watt (approved):
just minor suggestions!

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: 8386ed7

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: 8386ed7

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: 8386ed7

PR #213, opened 2023-05-16
Expose Publisher.asAnyWorkflow as public API

amorde:
Without this API consuming a `Publisher` inside a workflow is only possible by either rendering it directly via `.render(in:)` or by first calling `mapOutput`:

```swift
somePublisher
    .mapOutput { $0 } // workaround for not being able to call `asAnyWorkflow()`
    .onOutput { state, output in
         // ...
    }.rendered(in: context)
```

### DIFF
--- a/WorkflowCombine/Sources/PublisherWorkflow.swift
+++ b/WorkflowCombine/Sources/PublisherWorkflow.swift
@@ -21,17 +21,17 @@ import Foundation
 import Workflow
 
 struct PublisherWorkflow<WorkflowPublisher: Publisher>: Workflow where WorkflowPublisher.Failure == Never {
-    public typealias Output = WorkflowPublisher.Output
-    public typealias State = Void
-    public typealias Rendering = Void
+    typealias Output = WorkflowPublisher.Output
+    typealias State = Void
+    typealias Rendering = Void
 
     let publisher: WorkflowPublisher
 
-    public init(publisher: WorkflowPublisher) {
+    init(publisher: WorkflowPublisher) {
         self.publisher = publisher
     }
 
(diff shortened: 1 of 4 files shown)


WorkflowCombine/Sources/PublisherWorkflow.swift
```diff
@@ -21,17 +21,17 @@ import Foundation
 import Workflow
 
 struct PublisherWorkflow<WorkflowPublisher: Publisher>: Workflow where WorkflowPublisher.Failure == Never {
-    public typealias Output = WorkflowPublisher.Output
-    public typealias State = Void
-    public typealias Rendering = Void
```
amorde, line 26: doing some cleanup while here - `PublisherWorkflow` isn't `public` so these don't need to be either

mjohnson12:
Do we want to encourage using publishers directly in a workflow like this instead of using a worker?

jamieQ (approved):

mjohnson12 (approved):

amorde:
> Do we want to encourage using publishers directly in a workflow like this instead of using a worker?

Is there a way to use `PublisherWorker` directly that I'm missing? Since its not `public`, the only way to use it is to call `.render(in:)` which skips other steps along the way

mjohnson12:
> > Do we want to encourage using publishers directly in a workflow like this instead of using a worker?
> 
> Is there a way to use `PublisherWorkflow` directly that I'm missing? Since its not `public`, the only way to use it is to call `.render(in:)` which skips other steps along the way
You can do `Publisher.running(in)` which just calls `asAnyWorkflow` in the implementation so I'm fine with exposing `asAnyWorkflow` as public.

PR #214, opened 2023-05-17
[WorkflowTesting] Use CustomDump to better visualize failure messages

DJBen:
## What is it?

Better visualize the error message by using `CustomDump` in `WorkflowTesting`.

- Introduces dependency to [`CustomDump`](https://github.com/pointfreeco/swift-custom-dump) from `WorkflowTesting`, which is a lightweight opinion-less library that improves printing of any objects by leveraging `Mirror`s. 
- In `WorkflowTesting`, replaces all the `XCTAssertEquals` into `XCTAssertNoDifference`, which is the go-to method to compare the diff the objects.

## Why the change?
It has been a constant pain to reveal the differences of two very large structs when `XCTAssertEquals` fails, especially when the difference is minor. Some people resort to https://www.diffchecker.com/, or even subscribe to it to alleviate the issue. The manual copy and paste process is still a sink of engineering time.

Introducing the diffing visualization helps developers pinpoint the assertion failure in both local test sor CI pipelines, and speed up development in a long run.

*Before*:
```swift
("State(..... 1000+ characters omitted)") is not equal to ("State(..... 1000+ characters omitted)")
```
*After*:
```swift
XCTAssertNoDifference failed: …
State(
... (the method automatically omits overlapping states)
  − property: false
  + property: true
)
(First: −, Second: +)
            
```

## Impact of the added dependency
The added dependency is added into `WorkflowTesting`, which is meant for the tests modules to depend on. It does not contain any semantic-altering change in workflows.

If the consumer module already depends on `CustomDump`, it may cause a version conflict. Fortunately, I synced the dependency to `CustomDump` with Square's internal dependent version to avoid this issue.

## Checklist

- [x] Unit Tests
- [x] UI Tests (n/a)
- [x] Snapshot Tests (iOS only) (n/a)
- [x] I have made corresponding changes to the documentation (n/a)


WorkflowTesting.podspec
```diff
@@ -19,6 +19,7 @@ Pod::Spec.new do |s|
     s.source_files = 'WorkflowTesting/Sources/**/*.swift'
 
     s.dependency 'Workflow', "#{s.version}"
+    s.dependency 'CustomDump', '~> 0.6.1'
```
jamieQ, line 22: my guess is this is not going to work as a change to the public repo because this has been forked and hosted internally, but AFAIK does not exist in the public CocoaPods spec repo

square-tomb: https://github.com/square/workflow-swift/actions/runs/5006839279/jobs/8973201477?pr=214#step:5:8

DJBen: I came across this and it seems public https://github.com/Golface/CocoaPods-Composable-Architecture

dnkoutso: Its public in terms of a public GH repo is having a podspec. Its not public as in "published in the CocoaPods open source registry" (or trunk). 

You can always double check if a pod is actually published in the CocoaPods registry by visiting cocoapods.org and searching for a pods name. If it appears then its published.

You will not be able to do:
```ruby
pod 'CustomDump', :git => `https://github.com/Golface/CocoaPods-Composable-Architecture.git`
```

as CocoaPods does not allow that.

dnkoutso: @DJBen this is a pain...we'd have to ask for someone to publish this in the open source repo or we cannot depend on it. An alternative is to duplicate all of it inside Workflow which is meh. Trying to think of alternatives..........

DJBen: If I publish it to the trunk, will it solve the issue? Because SQ internally is using a fork, will that cause a version conflict?

dnkoutso: If you do it would solve this issue here I believe yes. 

For SQ if it uses an internal fork it would have to ensure the version requirements specified here match and if they do then I think it will take the fork

square-tomb (commented):
Thank you for contributing this! I share your expectation that just about everyone will prefer this output.

I think we could avoid introducing a new third-party dependency to workflow-swift by either:
-  adding a parameter to `tester` or `assert`, that accepts a function to use in place of `XCTAssertEqual`
- writing an extension within Register that adds a new set of `assert` methods

But both of these would require updating existing each `tester`/`assert` call site in Register that wants the new behavior, which we don't want to do if this is strictly an improvement. So I think making the change in WorkflowTesting itself is justified.

Can you put up a WIP PR in ios-register that points at this branch of workflow-swift, to verify with CI that none of our tests are somehow depending on specific behaviors of `XCTAssertEqual`?

WorkflowTesting.podspec
DJBen: I tried to untangle the public dependency and here's what I find:
- `CustomDump` depends on `XCTestDynamicOverlay`, but it needs > 0.4. Unfortunately, an old 0.1.1 is occupied by another person github.com/Arafa.

Is it okay that I alias the pods with `PFC` (pointfreeco), so they become `PFCXCTestDynamicOverlay` and `PFCCustomDump`, and so I can upload our versions to the public trunk?

DJBen: @gilgtm I managed to contact Rafa and persuade him to delete the `XCTestDynamicOverlay` repo so I can push a new version. I managed to push `CustomDump 0.11.0` to cocoapods.

jamieQ:
closing as i believe this was implemented via https://github.com/square/workflow-swift/pull/318

PR #216, opened 2023-05-22
[fix]: resolve memory exclusivity violations in some WorkflowObserver paths

jamieQ:
### Issue

- under some circumstances, the action application observation codepath could trigger runtime memory exclusivity exceptions

### Description

- added a test case that reproduces the issue with the existing API
- removed an unneeded `inout` parameter to resolve the problem

## Checklist

- [x] Unit Tests
- [x] UI Tests (N/A)
- [x] Snapshot Tests (iOS only) (N/A)
- [x] I have made corresponding changes to the documentation (N/A)

### DIFF
--- a/Workflow/Sources/WorkflowNode.swift
+++ b/Workflow/Sources/WorkflowNode.swift
@@ -84,7 +84,6 @@ final class WorkflowNode<WorkflowType: Workflow> {
             /// allowing the underlying conformance to be applied to the Workflow's State
             let outputEvent = openAndApply(
                 action,
-                to: &state,
                 isExternal: source == .external
             )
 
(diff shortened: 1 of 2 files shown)


Workflow/Sources/WorkflowNode.swift
```diff
@@ -182,13 +181,10 @@ private extension WorkflowNode {
     /// Applies an appropriate `WorkflowAction` to advance the underlying Workflow `State`
     /// - Parameters:
     ///   - action: The `WorkflowAction` to apply
-    ///   - state: The `State` to which the action will be applied
-    ///   - observerInfo: Optional observation info to notify registered `WorkflowObserver`s
     ///   - isExternal: Whether the handled action came from the 'outside world' vs being bubbled up from a child node
     /// - Returns: An optional `Output` produced by the action application
     func openAndApply<A: WorkflowAction>(
         _ action: A,
-        to state: inout WorkflowType.State,
```
jamieQ, line 191: the diff makes this a bit confusing, but removing this parameter just means that `self.state` is no longer shadowed in this method

amorde (approved):

square-tomb (approved):

PR #218, opened 2023-05-23
[feat]: merge ViewEnvironmentPropagatingObject into ViewEnvironmentPropagating

n8chur:
> **Note**:
> This PR is targeting the [`feature/viewenvironmentui`](https://github.com/square/workflow-swift/tree/feature/viewenvironmentui) branch.
> Reviewing this PR by commit may be ideal.

# Overview

## Merge `ViewEnvironmentPropagatingObject` into `ViewEnvironmentPropagating`

In practice, all graphs attempting to adopt these protocols will be objects—value type systems (like `WorkflowUI`'s propagation of `ViewEnvironment` through renderings) will not need to maintain nodes.

My adding the restriction to `AnyObject` on the root protocol, we can perform pointer comparisons of nodes (see below for why that's useful).

This consolidation of protocols likely also makes the system easier to understand.

## Only invalidate descendants with a matching ancestor

Updates the descendant invalidation check such that a node will only invalidate descendants if the descendant node's ancestor is set to that node. This prevents invalidating nodes that have their propagation paths overridden (e.g. `ViewControllerDescription`'s custom ancestor node). This should improve performance by reducing unnecessary invalidations.

# Square Integration PRs

https://github.com/squareup/market/pull/6286
https://github.com/squareup/ios-register/pull/86301

# Checklist

- [x] Unit Tests
- [x] UI Tests


ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -105,22 +126,248 @@ extension ViewEnvironmentPropagating {
...
+    /// Adds a `ViewEnvironment` change observation.
+    ///
+    /// The observation will only be active for as long as the returned lifetime is retained or
+    /// `cancel()` is called on it.
+    ///
+    @_spi(ViewEnvironmentWiring)
+    public func addEnvironmentNeedsUpdateObserver(
+        _ onNeedsUpdate: @escaping (ViewEnvironment) -> Void
+    ) -> ViewEnvironmentUpdateObservationLifetime {
+        let object = NSObject()
```
kyleve, line 172: Can we make this a specific type (eg, `NeedsEnvironmentUpdateToken` or something) so that it's easier to understand if/when we see these in the memory debugger?

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -105,22 +126,248 @@ extension ViewEnvironmentPropagating {
...
+    @_spi(ViewEnvironmentWiring)
+    public var environmentAncestorOverride: EnvironmentAncestorProvider? {
+        get {
+            objc_getAssociatedObject(self, &AssociatedKeys.ancestorOverride) as? EnvironmentAncestorProvider
+        }
+        set {
+            assert(
+                newValue == nil
+                    || environmentAncestorOverride == nil,
+                "Attempting to set environment ancestor override when one is already set."
```
kyleve, line 236: IMO it's a bit odd to require clearing a setter before you set it again. Thoughts?

kyleve (approved):

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: I'm still not clear on how this would help / when you'd run into this, but I don't have a strong take. Will update.

- [ ] Use special type for object tokens

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: Yup, it's definitely a bit weird, but we had it in place to ensure folks aren't overriding an existing override without knowing it—this assertion has caught unexpected scenarios like this in the past. 

Now that the framework has evolved a bit more, and I think we're more confident in the right shape for how to utilize these types, I don't feel as strongly that it's still necessary. I'm fine with removing it now if other folks agree!

robmaceachern: +1. Is it intended that we allow this to be `nil`'d out?

robmaceachern (approved):

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
watt: I'm in favor of leaving it. If the propagation path gets messed up it can cause very subtle bugs downstream that are difficult to diagnose. I think the extra step to unwire existing overrides before wiring new ones is worth it to avoid that kind of mistake.

watt (approved):

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -105,22 +126,248 @@ extension ViewEnvironmentPropagating {
...
+public final class ViewEnvironmentUpdateObservationLifetime {
+    /// Removes the observation.
+    ///
+    /// This is called in `deinit`.
+    ///
+    public func remove() {
+        onRemove()
+    }
+
+    var onRemove: () -> Void
```
square-tomb, line 298: Should this be `let` and/or `private`?

square-tomb (approved):

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -105,22 +126,248 @@ extension ViewEnvironmentPropagating {
...
+                self,
+                &AssociatedKeys.needsUpdateObservers
+            ) as? [NSObject: ViewEnvironmentUpdateObservation] ?? [:]
+        }
+        set {
+            objc_setAssociatedObject(
+                self,
+                &AssociatedKeys.needsUpdateObservers,
+                newValue,
+                .OBJC_ASSOCIATION_RETAIN
```
jamieQ, line 351: since this is UI code, should the nonatomic association be used? i assume this should never run concurrently

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: OK, I'm leaning towards keeping it for now, but let me know if anyone feels strongly that this shouldn't require unwiring before overriding!

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: I think you're right that this could/should be nonatomic! We should only be accessing this from the main actor/thread.

97b49ca

cc @watt

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: made it `private let`: eba72c6

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
watt: Seems OK to me! I don't recall any reason it was done this way originally.

PR #221, opened 2023-06-02
[feat]: add experimental WorkflowUI telemetry hooks

jamieQ:
### Motivation

we'd like to be able to build out some observation tooling that is aware of certain events that occur within WorkflowUI. specifically, view lifecycle data is of particular interest. currently the options for tying into UIViewController lifecycle events are either:

1. introduce new subclasses in client code that override the methods of interest and provide observation hooks there.
2. swizzle the methods of interest to allow client observation

option `1.` may place a large burden on consumers that have many existing subclasses (of, say, ScreenViewController), and only works for the `open` types defined in `WorkflowUI`, which excludes `WorkflowHostingController` & `DescribedViewController`. option `2.` may require less code to implement, but adds indirection (swizzling is often not particularly discoverable), runtime overhead, and may not work for Swift-only methods.

note: this is all currently implemented behind 'experimental' SPI declarations. the intent would be to treat this as sort of an 'alpha' version of this functionality, and to set expectations that it may have breaking changes going forward, without a major version increment in the library.

### Implementation

to achieve this goal, we've implemented an approach that does the following:

- unifies all `UIViewController` subtypes defined in `WorkflowUI` under a new parent class `WorkflowUIViewController`
- instruments some view controller lifecycle events (viewDidAppear, etc) within the shared base class
- exposes a static value that can be used to register a 'global' observer of these events
- adds some new protocols and event data types to enable observation of WorkflowUI lifecycle events

### Usage Example

the envisioned consumer is something like a shared telemetry system that wishes to observe certain WorkflowUI events and directly use or attach metadata to them. for example, imagine a consumer wished to log an event whenever a ViewController from WorkflowUI appeared on screen. with the proposed changes, this could be accomplished as follows:

```swift
// implement an observer
struct AppearanceLogger: WorkflowUIObserver {
    var loggingSystem: LoggingSystem

    // handle all emitted observation events
    func observeEvent<E: WorkflowUIEvent>(_ event: E) {
        // perform specific observation logic
        if let appearanceEvent = event as? ViewDidAppearEvent {
            loggingSystem.logAppearanceEvent(appearanceEvent.viewController)
        }
    }
}

// install the global observer
WorkflowUIObservation.sharedUIObserver = AppearanceLogger()
```

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
+++ b/WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
@@ -85,11 +85,6 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
         update(child: \.rootViewController, with: screen, in: environment)
 
         updatePreferredContentSizeIfNeeded()
-
-        sendObservationEvent(WorkflowHostingControllerDidUpdate(
-            hostingController: self,
-            newScreen: screen
-        ))
     }
 
     override public func viewDidLoad() {
@@ -153,25 +148,5 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
 
         preferredContentSize = newPreferredContentSize
     }
-}
-
-// MARK: Observation Events
-
-public protocol WorkflowHostingControllerEvent: WorkflowUIEvent {
-    associatedtype ScreenType: Screen
-    associatedtype Output
-
-    var hostingController: WorkflowHostingController<ScreenType, Output> { get }
-}
-
-extension WorkflowHostingControllerEvent {
-    public var viewController: UIViewController {
-        hostingController
-    }
-}
-
-public struct WorkflowHostingControllerDidUpdate<ScreenType: Screen, Output>: WorkflowHostingControllerEvent {
-    public let hostingController: WorkflowHostingController<ScreenType, Output>
-    public let newScreen: ScreenType
 }
 #endif
--- a/WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
+++ b/WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
@@ -18,13 +18,21 @@
 
 import ReactiveSwift
 import UIKit
+@_spi(ViewEnvironmentWiring) import ViewEnvironmentUI
 import Workflow
 
 /// Drives view controllers from a root Workflow.
 public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIViewController where ScreenType: Screen {
+    public typealias CustomizeEnvironment = (inout ViewEnvironment) -> Void
+
     /// Emits output events from the bound workflow.
     public var output: Signal<Output, Never> {
         return workflowHost.output
+    }
+
+    /// An environment customization that will be applied to the environment of the root screen.
+    public var customizeEnvironment: CustomizeEnvironment {
+        didSet { setNeedsEnvironmentUpdate() }
     }
 
     private(set) var rootViewController: UIViewController
@@ -33,15 +41,9 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
 
     private let (lifetime, token) = Lifetime.make()
 
-    public var rootViewEnvironment: ViewEnvironment {
-        didSet {
-            update(screen: workflowHost.rendering.value, environment: rootViewEnvironment)
-        }
-    }
-
     public init<W: AnyWorkflowConvertible>(
         workflow: W,
-        rootViewEnvironment: ViewEnvironment = .empty,
+        customizeEnvironment: @escaping CustomizeEnvironment = { _ in },
         observers: [WorkflowObserver] = []
     ) where W.Rendering == ScreenType, W.Output == Output {
         self.workflowHost = WorkflowHost(
@@ -49,14 +51,23 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
             observers: observers
         )
 
+        self.customizeEnvironment = customizeEnvironment
+
+        var customizedEnvironment: ViewEnvironment = .empty
+        customizeEnvironment(&customizedEnvironment)
+
         self.rootViewController = workflowHost
             .rendering
             .value
-            .buildViewController(in: rootViewEnvironment)
-
-        self.rootViewEnvironment = rootViewEnvironment
+            .viewControllerDescription(environment: customizedEnvironment)
+            .buildViewController()
 
         super.init(nibName: nil, bundle: nil)
+
+        // Do not automatically forward environment did change notifications to the rendered screen's backing view
+        // controller. Instead rely on `ViewControllerDescription` to call `setNeedsEnvironmentUpdate()` when updates
+        // occur.
+        environmentDescendantsOverride = { [] }
 
         addChild(rootViewController)
         rootViewController.didMove(toParent: self)
@@ -68,7 +79,7 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
             .observeValues { [weak self] screen in
                 guard let self = self else { return }
 
-                self.update(screen: screen, environment: self.rootViewEnvironment)
+                self.update(screen: screen, environment: self.environment)
             }
     }
 
@@ -82,7 +93,16 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
     }
 
     private func update(screen: ScreenType, environment: ViewEnvironment) {
+        let previousRoot = rootViewController
+
         update(child: \.rootViewController, with: screen, in: environment)
+
+        if previousRoot !== rootViewController {
+            // If a new view controller was instantiated and added as a child we need to inform it that the environment
+            // should be re-requested in order to respond to customizations in this WorkflowHostingController or any
+            // view controller above it in the UIViewController hierarchy.
+            setNeedsEnvironmentUpdate()
+        }
 
         updatePreferredContentSizeIfNeeded()
     }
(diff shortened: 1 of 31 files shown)


WorkflowUI/Sources/Observation/WorkflowUIEvents.swift
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+import Foundation
+import UIKit
+
+public protocol WorkflowUIEvent {
```
jamieQ, line 21: looking for feedback on the general structuring of the event types and related protocols

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
```diff
@@ -150,4 +155,23 @@ public final class WorkflowHostingController<ScreenType, Output>: UIViewControll
     }
 }
 
+// MARK: Observation Events
+
+public protocol WorkflowHostingControllerEvent: WorkflowUIEvent {
```
jamieQ, line 160: i'm not sure how useful this protocol is... since you have to know all the generic types up front, there isn't really a good way to do something like: 'given this `WorkflowUIEvent`, conditionally do something if it was emitted from a WorkflowHostingController'. you basically can't know that the event came from a hosting controller unless we were to add either some shared metadata to the base event protocol that specifies that, or make a marker protocol that ignores the associated types (or something else?).

WorkflowUI/Tests/WorkflowUIObservationTestCase.swift
```diff
@@ -0,0 +1,156 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Combine
+import Workflow
+@_spi(WorkflowUIGlobalObservation) import WorkflowUI
+import XCTest
+
+open class WorkflowUIObservationTestCase: XCTestCase {
```
jamieQ, line 22: there's some scratch work in here, but some ideas for how one could tie into this system and forward events through Combine

WorkflowUI/Sources/Observation/WorkflowUIEvents.swift
```diff
@@ -0,0 +1,48 @@
...
+    var viewController: UIViewController { get }
+}
+
+// MARK: ViewController Lifecycle Events
+
+public protocol ViewControllerLifeCycleEvent: WorkflowUIEvent {}
```
amorde, line 27: I see the intent but since `WorkflowUIEvent` always has a `UIViewController`, does this add any value?

WorkflowUI/Sources/Observation/WorkflowUIViewController.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        sendObservationEvent(ViewDidAppearEvent(
+            viewController: self,
+            animated: animated,
+            isFirstAppearance: isFirstAppearance
+        ))
+    }
+
+    override open func viewWillLayoutSubviews() {
+        // no need to call super since it does nothing
```
amorde, line 61: technically correct, though it does feel a bit wrong..

amorde (approved):

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
amorde: Agree, this doesn't seem to add value other than enforcing some naming conventions for the generic types

WorkflowUI/Sources/Observation/WorkflowUIEvents.swift
amorde: Structure looks good to me

WorkflowUI/Sources/Observation/WorkflowUIViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#if canImport(UIKit)
+import Foundation
+import UIKit
+
+/// Ancestor type from which all ViewControllers in WorkflowUI inherit.
+open class WorkflowUIViewController: UIViewController {
```
jamieQ, line 22: cc @n8chur i know you've got a feature branch that this might conflict with. any issues with adding this new common supertype you can think of?

mjohnson12 (approved):
This looks good to me and flows well with the rest of the WorkflowUI code.

Kiesco08 (approved):
Approach lgtm too, seems easy to use

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
```diff
@@ -99,7 +99,7 @@ public final class WorkflowHostingController<ScreenType, Output>: UIViewControll
     }
 
     override public func viewDidLayoutSubviews() {
-        super.viewDidLayoutSubviews()
+        defer { super.viewDidLayoutSubviews() }
```
watt, line 102: Is this change needed for your hooks to work? We were considering moving the frame assignment to `viewWillLayoutSubviews` as part of the `ViewEnvironmentUI` feature, but wanted to isolate that change due to potential for side effects in code that was depending on this order of events.

square-tomb (approved):
Looks good to me!

WorkflowSwiftUI contains a type `WorkflowHostingViewController` that should probably now subclass `WorkflowUIViewController`, too.

Other approaches to SwiftUI support that we're experimenting with involve [a view controller type that sublclasses `UIHostingController`](https://github.com/square/workflow-swift/blob/93c3b5d281bff9de2e40390c1192a25dfe8a7a45/WorkflowSwiftUI/Sources/SwiftUIScreen.swift#L83) and so cannot subclass `WorkflowUIViewController`.

However, I think any view controller type like that can be modified to contain, rather than subclass, `UIHostingController`. The only downside AFAIK would be a deeper VC hierarchy.

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
jamieQ: remove this for now, and may revisit in the future

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
jamieQ: no i don't think this change is necessary. the previous implementation of `super.viewDidLayoutSubviews()` did nothing (according to the docs), and the new one emits an event. it seems like the event emission should occur after any logic in this method, but i don't think the ordering relative to the frame assignment will matter to any of the envisioned consumers.

if we're potentially going to move the frame assignment out of this method, then perhaps we should leave this proposed change since it will preserve the relative ordering with the new observation events?

WorkflowUI/Tests/WorkflowUIViewControllerTests.swift
```diff
@@ -0,0 +1,239 @@
...
+    var value: Int = 0
+
+    func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
+        .testing
+    }
+}
+
+private extension ViewControllerDescription {
+    static var testing: Self {
+        .init(build: { UIViewController() }, update: { _ in })
```
n8chur, line 236: I think this would be the only conflict when compared to the ViewEnvironmentUI work—`ViewControllerDescription` requires an `environment` parameter in that feature branch. It should be easy to reconcile depending on who lands their work first!

n8chur (approved):
I'm a bit nervous about the reliance on a `UIViewController` superclass when `ViewControllerDescription` does not enforce this, but it seems like it may be worth the alternative given our goals!

`ViewControllerDescriptions` not describing a `ScreenViewController` subclass does seem like a potential footgun here (what @square-tomb pointed out in a previous comment).

Are we planning on adding documentation that describes in detail how to adhere to these event contract rules once it's out of alpha? Could Swift macros help make adding adherence easier in the future without needing to wrap one VC in another? It'd be nice to avoid needing to compose a nested VC to get this behavior.

kyleve:
> ViewControllerDescriptions not describing a ScreenViewController subclass does seem like a potential footgun here

Agreed. I don't think this is something we can/should enforce here, IMO.

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
```diff
@@ -22,7 +22,7 @@ import UIKit
 import Workflow
 
 /// Drives view controllers from a root Workflow.
-public final class WorkflowHostingController<ScreenType, Output>: UIViewController where ScreenType: Screen {
+public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIViewController where ScreenType: Screen {
```
jamieQ, line 25: @square-tomb responding here for threading purposes:

> WorkflowSwiftUI contains a type `WorkflowHostingViewController` that should probably now subclass `WorkflowUIViewController`, too.

good callouts, but given the current motivations for this interface, i think it's okay to punt on the story with SwiftUI and ignore this for the time being.

jamieQ, line 25: @kyleve responding here for threading purposes:

> >ViewControllerDescriptions not describing a ScreenViewController subclass does seem like a potential footgun here
> 
> Agreed. I don't think this is something we can/should enforce here, IMO.

could you clarify – you think this is a risk, but not one we should change the contract of ViewControllerDescription to deal with?

jamieQ: > I wonder, do we need events from`ScreenViewController` at all? Maybe the events from `DescribedViewController` are sufficient if that VC usually maps 1:1 to a screen.

being able to recover the Screen type from ScreenViewController may be of use

> If not, and `DescribedViewController` is often used to display a succession of different screens, we would have to add some events [here](https://github.com/square/workflow-swift/pull/221/files#diff-20be64c4e0d93842089c1503234a2662af08e4a919c9f946e737e94015f5e47bR45-R47). Or we could wrap `currentViewController` in a container `WorkflowUIViewController` if we don't mind deepening the VC hierarchy.

adding some additional events that are more specific to the logic in some of the VC subtypes may be something we want to do eventually, but for now i thought we'd just start with VC lifecycle information and see how far that gets us.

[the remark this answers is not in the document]
n8chur: > i thought macros had to be additive, so i'm not sure we'd be able to change any existing code in client-supplied VCs.

I dove a little deeper into Swift Macros after my initial comment and I think you're right, it looks like it may not be useful here.

PR #222, opened 2023-06-07
Remove redundant `canImport` conditions

square-tomb:
Prior to this PR, every declaration in WorkflowUI is wrapped in `#if canImport(UIKit)`. This allows a platform without UIKit (such as macOS) to import WorkflowUI, but doing so would be pointless as it would import zero symbols.

Likewise with WorkflowCombine and `#if canImport(UIKit)`.

This PR removes all of those `canImport` conditions. This is technically a breaking change for hypothetical clients that been importing those frameworks without receiving symbols from them. Those clients should delete those `import` statements.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowCombine/Sources/Publisher+Extensions.swift
+++ b/WorkflowCombine/Sources/Publisher+Extensions.swift
@@ -5,8 +5,6 @@
 //  Created by Soo Rin Park on 11/3/21.
 //
 
-#if canImport(Combine)
-
 import Combine
 import Foundation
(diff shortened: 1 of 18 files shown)


square-tomb:
Turns out that SPM doesn't give us any way to specify that WorkflowUI target is not available for the macOS platform. Therefore the WorkflowUI target is still built for macOS, which build fails when UIKit cannot be imported.

PR #227, opened 2023-06-28
[fix]: update WorkflowHostingController on layout if ancestor hierarchy has changed

n8chur:
# Problem

When a `WorkflowHostingController` is added to a `UIViewController` hierarchy, we need to ensure that we re-render the `Screen` using an updated `ViewEnvironment`, since the environment from the ancestor path above that `WorkflowHostingController` could contain customizations that are not present in the `WorkflowHostingController` itself.

Currently, it is expected that consumers would call `setNeedsEnvironmentUpdate()` when adding a `WorkflowHostingController` as a child, but this is fragile and hard to enforce/validate, especially in legacy code.

Omitting this `setNeedsEnvironmentUpdate()` can cause issues where content would be rendered on screen using invalid `ViewEnvironment` values on the first rendering. On a later environment update, we'd likely see this environment reflect the correct values, but we want to avoid showing content with bad `ViewEnvironment` information whenever possible. An example of this issue can be seen in [this Slack thread from a in issue reported in `ios-register`](https://square.slack.com/archives/CTJ4UNLHF/p1687895879847009).

# Solution

While consumers should still try to call `setNeedsEnvironmentUpdate()` whenever the environment changes (either through an update in value or an update in path), this PR attempts to catch mistakes and legacy usecases by tracking the ancestor tree node path when state updates and layouts occur, and if that path changes, we'll ensure an update occurs on the next layout pass.

# Integration

I've smoke tested this change in `ios-register` and `market` and no changes are needed upstream to support this fix. Here are the integration PRs (with no real code changes) that show passing CI jobs:

`ios-register`: https://github.com/squareup/ios-register/pull/88056
`market`: https://github.com/squareup/market/pull/6547

### DIFF
--- a/ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
+++ b/ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
@@ -272,6 +272,45 @@ extension ViewEnvironmentPropagating {
             objc_setAssociatedObject(self, &AssociatedKeys.descendantsOverride, newValue, .OBJC_ASSOCIATION_RETAIN)
         }
     }
+
+    /// Returns an `Equatable` representation of the `ViewEnvironmentPropagating` environment
+    /// ancestor tree path.
+    ///
+    /// This can be useful, for example, if you need to determine if any ancestor was inserted or
+    /// removed above this node.
+    ///
+    /// The `Equatable` implementation of this type compares the tree as an array of object
+    /// identifiers for each node.
+    ///
+    @_spi(ViewEnvironmentWiring)
+    public var environmentAncestorPath: EnvironmentAncestorPath {
+        var path = EnvironmentAncestorPath()
+
+        if let first = environmentAncestor {
+            for node in sequence(first: first, next: \.environmentAncestor) {
+                path.nodes.append(ObjectIdentifier(node))
+            }
+        }
+
+        return path
+    }
+
+    @_spi(ViewEnvironmentWiring)
+    public typealias EnvironmentAncestorPath = ViewEnvironmentPropagatingAncestorPath
+}
+
+/// An `Equatable` representation of the `ViewEnvironmentPropagating` environment ancestor tree
+/// path.
+///
+/// This can be useful, for example, if you need to determine if any ancestor was inserted or
+/// removed above this node.
+///
+/// The `Equatable` implementation of this type compares the tree as an array of object
+/// identifiers for each node.
+///
+@_spi(ViewEnvironmentWiring)
+public struct ViewEnvironmentPropagatingAncestorPath: Equatable {
+    var nodes: [ObjectIdentifier] = []
 }
 
 /// A closure that is called when the `ViewEnvironment` needs to be updated.
--- a/ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
+++ b/ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
@@ -279,8 +279,8 @@ extension ViewEnvironmentPropagating {
     /// This can be useful, for example, if you need to determine if any ancestor was inserted or
     /// removed above this node.
     ///
-    /// The `Equatable` implementation of this type compares the tree as an array of object
-    /// identifiers for each node.
+    /// The `Equatable` implementation of this type compares the tree as an array of weak
+    /// references.
     ///
     @_spi(ViewEnvironmentWiring)
     public var environmentAncestorPath: EnvironmentAncestorPath {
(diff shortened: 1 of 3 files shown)


ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -272,6 +272,20 @@ extension ViewEnvironmentPropagating {
             objc_setAssociatedObject(self, &AssociatedKeys.descendantsOverride, newValue, .OBJC_ASSOCIATION_RETAIN)
         }
     }
+
+    @_spi(ViewEnvironmentWiring)
+    public var environmentAncestorTreeHash: AnyHashable {
```
kyleve, line 277: Maybe this is just for the prototype; but the hash isn't guaranteed to be different if the tree is different. I'd throw this into something like:

```
struct Ancestors : Equatable {

  private var IDs : [ObjectIdentifier]

  init(node : ViewController) {
    IDs = sequence(first: first, next: \.environmentAncestor).map {
      ObjectIdentifier($0)
    }
  }
}
```

n8chur: `ObjectIdentifier` is what I was looking for, thanks!
But yeah, I knew this wasn't quite right.

mjohnson12 (approved):

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
```diff
@@ -120,7 +128,14 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
 
     override public func viewWillLayoutSubviews() {
         super.viewWillLayoutSubviews()
-        applyEnvironmentIfNeeded()
+
+        let environmentAncestorPath = environmentAncestorPath
+        if environmentAncestorPath != lastEnvironmentAncestorPath {
```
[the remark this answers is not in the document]
n8chur: It could be become root of a window where there would be no ancestor and I think we'd still want to perform an update in that case.

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
```diff
@@ -41,6 +41,8 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
 
     private let (lifetime, token) = Lifetime.make()
 
+    private var lastEnvironmentAncestorPath: EnvironmentAncestorPath?
```
watt, line 44: I think we might have a potential retain cycle here, if this is storing strong references to the entire VC hierarchy above it. Can we weak-box each ancestor on the path, or does that break our ability to reliably compare paths?

kyleve: It's not a reference since it uses `ObjectIdentifier(node)` right? Eg, just a pointer.

nononoah: That's also my understanding.

n8chur: I talked through this with @watt and did some experimentation locally and it does look like `ObjectIdentifier` just keeps a reference to the pointer value, not the object itself.

One interesting note that @watt pointed out from [the documentation of `ObjectIdentifier`](https://github.com/apple/swift/blob/main/stdlib/public/core/ObjectIdentifier.swift#L15-L16):
> This unique identifier is only valid for comparisons during the lifetime of the instance.

Andrew was wondering if the pointer could be reused for a new instance and cause false positives on the equitability check.

This seems rare and very unlikely to occur, but we could add a weak box for these objects to better track the lifetime if we think it's worth the effort.

n8chur: I've updated this to use a weak box and `===` comparisons: fa85e0f

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -272,6 +272,66 @@ extension ViewEnvironmentPropagating {
...
+}
+
+/// An `Equatable` representation of the `ViewEnvironmentPropagating` environment ancestor tree
+/// path.
+///
+/// This can be useful, for example, if you need to determine if any ancestor was inserted or
+/// removed above this node.
+///
+/// The `Equatable` implementation of this type compares the tree as an array of object
+/// identifiers for each node.
```
watt, line 309:
```suggestion
/// The `Equatable` implementation of this type compares the tree as an array of weak references.
```

watt (approved):

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: c1dd4a3

kyleve (approved):

PR #231, opened 2023-07-13
[WIP DNR] [feat]: add imperative environment customization support

n8chur:
## Checklist

- [ ] Unit Tests
- [ ] UI Tests


ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -432,3 +462,62 @@ extension ViewEnvironmentPropagating {
...
+public final class ViewEnvironmentCustomizationLifetime {
+    /// Removes the observation.
+    ///
+    /// This is called in `deinit`.
+    ///
+    public func remove() {
+        onRemove()
+    }
+
+    private let onRemove: () -> Void
```
watt, line 487: In `TrampolineModalPresenter.LifetimeToken` we nil out the callback after the first call, and assert if you try to manually dismiss twice. What do you think about doing that here too?

ViewEnvironmentUI/Tests/ViewEnvironmentObservingTests.swift
```diff
@@ -324,6 +324,70 @@ final class ViewEnvironmentObservingTests: XCTestCase {
...
+        // Customizations should be respected.
+        do {
+            var customizationLifetime: ViewEnvironmentCustomizationLifetime? = viewController
+                .addEnvironmentCustomization {
+                    $0.testContext.number = 200
+                }
+
+            XCTAssertEqual(viewController.environment.testContext.number, 200)
+
+            _ = customizationLifetime // Suppress warning about variable never being read
```
watt, line 342: Wrapping up the lifetime's scope in `withExtendedLifetime` is another way.

ViewEnvironmentUI/Tests/ViewEnvironmentObservingTests.swift
```diff
@@ -324,6 +324,70 @@ final class ViewEnvironmentObservingTests: XCTestCase {
...
+            customizationLifetime = nil
+
+            // Customization should be removed when lifetime is deallocated
```
watt, line 345: You might also want to test explicitly calling `remove()`

watt (approved):

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: I don't have a strong take—I'm fine with adding it!

- [ ] assert `remove()` is only called at most once

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -176,6 +180,31 @@ extension ViewEnvironmentPropagating {
...
     }
 
+    /// Adds a `ViewEnvironment` customization to this node.
+    ///
+    /// These customizations will occur before the node's `customize(environment:)` in cases where
+    /// this node conforms to `ViewEnvironmentObserving`, and will occur the order in which they
+    /// were added.
+    ///
+    /// The customization will only be active for as long as the returned lifetime is retained or
+    /// `cancel()` is called on it.
```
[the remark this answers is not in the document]
n8chur: - [ ] Fix typo in doc comment (`cancel()` -> `remove()`)

ViewEnvironmentUI/Tests/ViewEnvironmentObservingTests.swift
n8chur: Sure!

- [ ] add explicit `remove()` call test

ViewEnvironmentUI/Tests/ViewEnvironmentObservingTests.swift
n8chur: Agree and I do prefer that, but this is the pattern used in the other tests in this module so I did the same here. I can migrate them all in this PR if you think it's worth the slightly increased scope—let me know!

n8chur: Nevermind, I'll just update these now, there's only 1!

n8chur: a693580f

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -351,17 +380,21 @@ public final class ViewEnvironmentUpdateObservationLifetime {
     /// This is called in `deinit`.
```
jamieQ, line 380: is this still true?

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -351,17 +380,21 @@ public final class ViewEnvironmentUpdateObservationLifetime {
...
     ///
     public func remove() {
+        guard let onRemove else {
+            preconditionFailure("Environment customization was already removed")
```
jamieQ, line 384: this seems like an assert warning about unexpected behavior may be more appropriate. this is idempotent, right (other than the precondition)? will something break horribly if it's called more than once?

n8chur: Nothing would break if this was called more than once with the current implementation.

I'm fine with moving this to an assert with early return, but I'll check with @watt who suggested this change!

- [ ] Use `assert` instead of `preconditionFailure`?

watt: Asserting is fine with me. I think the idea here is that we consider it a programmer error to try to remove a customization more than once, even if it is effectively idempotent.

n8chur: Oh whoops, I added this to the wrong lifetime 🤦 

- [ ] Apply removal assertion changes to customization lifetime

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -176,6 +180,31 @@ extension ViewEnvironmentPropagating {
...
+    ///
+    /// These customizations will occur before the node's `customize(environment:)` in cases where
+    /// this node conforms to `ViewEnvironmentObserving`, and will occur the order in which they
+    /// were added.
+    ///
+    /// The customization will only be active for as long as the returned lifetime is retained or
+    /// until `remove()` is called on it.
+    ///
+    @_spi(ViewEnvironmentWiring)
+    public func addEnvironmentCustomization(
```
jamieQ, line 193: just to better my own understanding – should this trigger the 'setNeedsEnvironmentUpdate' (or whatever its name is) logic, or does that responsibility lie at a different level?

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
```diff
@@ -119,6 +119,10 @@ extension ViewEnvironmentPropagating {
     public var environment: ViewEnvironment {
```
jamieQ, line 119: more of a general question about the changeset – is this alternate API mainly for convenience? curious what the motivating examples are (if they exist).

jamieQ (approved):
this all looks reasonable to me!

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: Yeah—I'm fairly torn on whether or not this should automatically call `setNeedsEnvironmentUpdate()` on your behalf. You'll almost always want to call it after adding a customization, but I figured there may be cases where you'd prefer to defer that call or manually manage when the invalidation occurs (e.g. for performance reasons). `setNeedsEnvironmentUpdate()` does trigger `environmentDidChange()` calls which are not deferred to the next layout cycle, and we use this to perform updates in some cases within WorkflowUI and Market. For example, WorkflowHostingController will re-render its content whenever `environmentDidChange()` is called.

The current shape also makes it a bit easier to ensure events happen at more appropriate times. For example, if you were to be adding a customization to a new child VC in a container VC, you might have something like this:

```swift
addChild(viewController)
customizationLifetime = viewController
    .addEnvironmentCustomization(makeCustomization(for: column))
if let view = viewIfLoaded {
    view.add(viewController.view)
}
viewController.didMove(toParent: self)
viewController.setNeedsEnvironmentUpdate()
```
 This order of operations means:
* The customization and parent/child relationship is established before the child VC's view is (conditionally) loaded. This ensures any `environment` access during `viewDidLoad` would have the appropriate environment wiring path and customizations.
* `setNeedsEnvironmentUpdate()` is called after the parent/child relationship is fully established (`didMove(toParent:)` has been called). I'm not aware of any real issues that could come from calling this before `didMove(toParent:)`, but it seems like the most appropriate time in the VC lifecycle IMO.

I'm very open to automatically calling `setNeedsEnvironmentUpdate()` automatically when a customization is added if we prefer the safety here!

ViewEnvironmentUI/Sources/ViewEnvironmentPropagating.swift
n8chur: It's mainly for convenience. The motivating example is support for what we call "pane-based size class", where a split container VC is able to override the horizontal size class for a particular pane based on the available horizontal width of that pane.

Without this change we'd either need to insert a node between the split container VC and it's children, or introduce a wrapper VC around those children to perform the environment mutation.

Issue #235, opened 2023-08-10
RenderTester's `expect(publisher: ...)` method should accept `nil` output


jamieQ:
currently you cannot straightforwardly use RenderTester to test a Publisher that does not emit any output during a render pass

PR #236, opened 2023-08-10
[fix]: support nil output in RenderTester Publisher extension

jamieQ:
### Issue

- the `WorkflowCombine` `RenderTester` extensions didn't allow `nil` output

### Description

- add a new method to support `nil` output and match the naming convention of analogous methods
- deprecate old method
- update tests

addresses https://github.com/square/workflow-swift/issues/235

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowCombine/Testing/PublisherTesting.swift
+++ b/WorkflowCombine/Testing/PublisherTesting.swift
@@ -23,11 +23,12 @@ import XCTest
 @testable import WorkflowCombine
 
 extension RenderTester {
-    /// Expect a `Publisher`s.
+    /// Expect a `Publisher`-based Workflow.
     ///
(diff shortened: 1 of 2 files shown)


WorkflowCombine/Testing/PublisherTesting.swift
```diff
@@ -43,6 +44,19 @@ extension RenderTester {
             assertions: { _ in }
         )
     }
+
+    @available(*, deprecated, renamed: "expect(publisher:producingOutput:key:)")
```
jamieQ, line 48: there are a handful of uses in the monorepo that can be migrated to avoid this warning

mjohnson12 (approved):

PR #238, opened 2023-08-11
Exhaustive RenderTester state assertion

gilgtm:
- Before: To perform exhaustive state assertions with a RenderTester you have to manually keep track of the initial state.
- After: Added RenderTesterResult/assertState which makes it easy to perform exhaustive state assertions by mutating the initial state

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowTesting/Sources/RenderTesterResult.swift
+++ b/WorkflowTesting/Sources/RenderTesterResult.swift
@@ -20,11 +20,13 @@ import XCTest
 /// The result of a `RenderTester` rendering. Used to verify state, output, and actions that were produced as a result of
 /// actions performed during the render (such as child workflow output being produced).
 public struct RenderTesterResult<WorkflowType: Workflow> {
+    let initialState: WorkflowType.State
     let state: WorkflowType.State
     let appliedAction: AppliedAction<WorkflowType>?
     let output: WorkflowType.Output?
 
-    internal init(state: WorkflowType.State, appliedAction: AppliedAction<WorkflowType>?, output: WorkflowType.Output?) {
+    internal init(initialState: WorkflowType.State, state: WorkflowType.State, appliedAction: AppliedAction<WorkflowType>?, output: WorkflowType.Output?) {
+        self.initialState = initialState
         self.state = state
         self.appliedAction = appliedAction
         self.output = output
@@ -38,6 +40,22 @@ public struct RenderTesterResult<WorkflowType: Workflow> {
         assertions: (WorkflowType.State) throws -> Void
     ) rethrows -> RenderTesterResult<WorkflowType> {
         try assertions(state)
+        return self
+    }
+
+    /// Exhaustive state testing against the initial state.
+    /// - Parameters:
+    ///   - changes: A function that receives the initial state
+    ///   and is expected to mutate it to match the new state.
+    @discardableResult
+    public func assertState(
+        file: StaticString = #file,
+        line: UInt = #line,
+        changes: (inout WorkflowType.State) throws -> Void
+    ) rethrows -> RenderTesterResult<WorkflowType> where WorkflowType.State: Equatable {
+        var initialState = self.initialState
+        try changes(&initialState)
+        XCTAssertEqual(initialState, state, "Initial state did not match new state", file: file, line: line)
         return self
     }
 
--- a/WorkflowTesting/Sources/RenderTesterResult.swift
+++ b/WorkflowTesting/Sources/RenderTesterResult.swift
@@ -43,22 +43,6 @@ public struct RenderTesterResult<WorkflowType: Workflow> {
         return self
     }
 
-    /// Exhaustive state testing against the initial state.
-    /// - Parameters:
-    ///   - changes: A function that receives the initial state
-    ///   and is expected to mutate it to match the new state.
-    @discardableResult
-    public func assertState(
-        file: StaticString = #file,
-        line: UInt = #line,
-        changes: (inout WorkflowType.State) throws -> Void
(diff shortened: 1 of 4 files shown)


WorkflowTesting/Sources/RenderTesterResult.swift
```diff
@@ -41,6 +43,22 @@ public struct RenderTesterResult<WorkflowType: Workflow> {
...
+    /// Exhaustive state testing against the initial state.
+    /// - Parameters:
+    ///   - changes: A function that receives the initial state
+    ///   and is expected to mutate it to match the new state.
+    @discardableResult
+    public func assertState(
+        file: StaticString = #file,
+        line: UInt = #line,
+        changes: (inout WorkflowType.State) throws -> Void
+    ) rethrows -> RenderTesterResult<WorkflowType> where WorkflowType.State: Equatable {
```
jamieQ, line 55: nit: there is an extension below (~L130) where the utilities for equatable states should probably all go

WorkflowTesting/Sources/RenderTesterResult.swift
```diff
@@ -41,6 +43,22 @@ public struct RenderTesterResult<WorkflowType: Workflow> {
...
+        var initialState = self.initialState
+        try changes(&initialState)
+        XCTAssertEqual(initialState, state, "Initial state did not match new state", file: file, line: line)
```
jamieQ, line 58: this error message seems a little confusing to me. we don't actually expect the 'initial state' to match the new state right? we expect the explicitly transformed state to match. not sure what the best wording to communicate this is though...

WorkflowTesting/Sources/RenderTesterResult.swift
```diff
@@ -41,6 +43,22 @@ public struct RenderTesterResult<WorkflowType: Workflow> {
         return self
     }
 
+    /// Exhaustive state testing against the initial state.
+    /// - Parameters:
+    ///   - changes: A function that receives the initial state
+    ///   and is expected to mutate it to match the new state.
+    @discardableResult
+    public func assertState(
```
jamieQ, line 51: i wonder if naming like `assertStateModifications`, `assertModifiedState`, `assertTransformedState`, or something along those lines might be a bit more clear. since there's already an `assert(state:)` method, might be good to make the difference a bit more obvious.

gilgtm: `assertStateModifications` is pretty close to what I have now if you read the argument as part of the name

```
assertState(changes:)
```

gilgtm: come to think of it `assert(stateChanges:)` or `assert(stateModifications:)` might be a closer match to the existing pattern. wdyt?

WorkflowTesting/Sources/RenderTesterResult.swift
gilgtm: - [ ] move to extension

WorkflowTesting/Sources/RenderTesterResult.swift
gilgtm: hmm yeah this doesn't make sense... perhaps something more vague like 

> Expected state does not match

gilgtm: also perhaps I should reverse the order here

```
XCTAssertEqual(state, initialState
```

WorkflowTesting/Tests/WorkflowRenderTesterFailureTests.swift
```diff
@@ -340,6 +340,18 @@ final class WorkflowRenderTesterFailureTests: XCTestCase {
             }
         }
     }
+
+    func test_assertState() {
+        let result = TestWorkflow()
+            .renderTester(initialState: .idle)
+            .render { _ in }
+
```
square-tomb, line 348: Shall we test that `assertState` can succeed as well?

I'm not sure why we haven't done so for the other, existing assert/verify methods 🤔

square-tomb (approved):
Thanks for this improvement. Looks good to me, and I agree with the naming/phrasing tweaks that you and @jamieQ are discussing.

This reminded me of [a post I read](https://www.merowing.info/exhaustive-testing-in-tca/) arguing against exhaustive testing in TCA. But re-reading the post, it really only makes a good case against _mandating all_ tests be exhaustive over state and effects.

WorkflowTesting/Sources/RenderTesterResult.swift
jamieQ: i guess my personal preference is just to try and avoid having a number of different methods that end up reading like `assert { ... }` with different closure types, as i find that harder to search for, and harder to understand when visually scanning.

WorkflowTesting/Tests/WorkflowRenderTesterFailureTests.swift
gilgtm: We are already doing that in `WorkflowRenderTesterTests/test_childWorkflowOutput` for both this new method and `verifyState`

WorkflowTesting/Sources/RenderTesterResult.swift
```diff
@@ -120,6 +122,22 @@ extension RenderTesterResult where WorkflowType.State: Equatable {
         XCTAssertEqual(state, expectedState, file: file, line: line)
         return self
     }
+
+    /// Exhaustive state testing against the initial state.
+    /// - Parameters:
+    ///   - changes: A function that receives the initial state
```
jamieQ, line 128: nit: param name doesn't match

jamieQ (approved):

PR #249, opened 2023-10-12
[feat]: external logging API

jamieQ:
## Checklist

- [ ] Unit Tests
- [ ] UI Tests


Workflow/Sources/ExternalLogging.swift
```diff
@@ -0,0 +1,70 @@
+import Foundation
+
+/// Namespace for logging API used to propagate internal Workflow-related logging to external consumers
+public enum ExternalLogging {}
```
mjohnson12, line 4: Love the use of the name space.

mjohnson12 (approved):
I like the approach and the items chosen to log.

attoh-square (commented):
I think for what we are trying to achieve here, the simplicity of this approach make it a good fit. Its nice that we don't have to add another dependency and this looks simpler to maintain where the existing telemetry infra may not be (it's makes sense to separate telemetry from general logging anyways).

> Clients are expected to configure this early on when bootstrapping.

Does this mean that nested workflows need to configure logging as well or configuring at the root workflow should be enough?

PR #253, opened 2023-10-30
[WIP] WorkflowSwiftUI without ObservedObject

square-tomb:
Experiment:
- No `SwiftUIScreen` protocol. SwiftUI View conforms directly to `Screen`
- No `ObservableValue`. SwiftUI View has multiple, plain, value-typed properties instead of a single `@ObservedObject` model
- Each Workflow re-render  sets `UIHostingController.rootView`

This approach previously resulted in animation issues documented in [Binding SwiftUI Views to Workflow Renderings](https://docs.google.com/document/d/1DemHF9RViaNBUkp4T3rPqigGK2pZBJqtsyVVudncZCY/edit#heading=h.a4fs99c5owvq) and [2020's WorkflowView](https://github.com/square/workflow-swift/blob/4399ddc1c588881b2871603e83b6ab5f4303f219/WorkflowSwiftUI/Sources/WorkflowView.swift#L270-L272), but I saw none in the SwiftUITestbed app under Xcode 14 and iOS 16.

This approach might lead to more view body evaluations, but after my performance noodling last month, I think those evaluations are cheap as long as they don't trigger additional work in the Core Animation commit phase.


WorkflowSwiftUIExperimental/Sources/Screen+View.swift
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#if canImport(UIKit)
+
+import SwiftUI
+import Workflow
+import WorkflowUI
+
+public extension Screen where Self: View {
+    func viewControllerDescription(environment: ViewEnvironment) -> ViewControllerDescription {
```
kyleve, line 24: Clever!

square-tomb:
> This approach previously resulted in animation issues documented in [Binding SwiftUI Views to Workflow Renderings - Google Docs](https://docs.google.com/document/d/1DemHF9RViaNBUkp4T3rPqigGK2pZBJqtsyVVudncZCY/edit#heading=h.a4fs99c5owvq) and [2020's WorkflowView](https://github.com/square/workflow-swift/blob/4399ddc1c588881b2871603e83b6ab5f4303f219/WorkflowSwiftUI/Sources/WorkflowView.swift#L270-L272), but I saw none in the SwiftUITestbed app under Xcode 14 and iOS 16.

Update: @mjohnson12 tested under iOS 15 and found the same broken animation for `Toggle` that we reported in that design doc.

https://github.com/square/workflow-swift/assets/106167956/0efbee20-7191-4e61-84ef-4741436f4ed0

Binding+Buffered.swift
square-tomb, line 1: This is possible local workaround for the problem of a Swift UI View accessing the getter of a stale `Binding`, which [it turns out](https://github.com/square/workflow-swift/pull/253#issuecomment-1787933874) `Toggle` still does in iOS 15.

(cc @mjohnson12 @n8chur)

mjohnson12: This file is showing in the root of the repo for me and the SwiftUITestbed doesn't build for me.

square-tomb: Oops. Fixed in 9f3c49dd.

PR #254, opened 2023-10-30
Equatable sinks prototype

watt:


Samples/SwiftUITestbed/Sources/MainScreen.swift
```diff
@@ -108,9 +103,26 @@ extension MainScreen: MarketBackStackContentScreen {
...
+    enum Action {
+        case pushScreen
+        case presentScreen
+        case changeTitle(String)
+        case changeAllCaps(Bool)
+        case close
+    }
+}
+
+// I guess this could be upstreamed to Blueprint
```
kyleve, line 123: Oh yeah, I think I've seen this go by in POS

Samples/SwiftUITestbed/Sources/MainWorkflow.swift
```diff
@@ -39,53 +41,96 @@ struct MainWorkflow: Workflow {
...
+
+    func makeSink<Action, WorkflowType>(
+        of actionType: Action.Type,
+        with context: RenderContext<WorkflowType>
+    ) -> StableSink<Action> where Action: WorkflowAction, Action.WorkflowType == WorkflowType {
+        let sink = context.makeSink(of: actionType)
+
+        sinks[ObjectIdentifier(actionType)] = sink
+
+        return StableSink(trampoline: self)
```
square-tomb, line 101: I wonder if we could pass `sink` into `StableSink.init` here and implement `StableSink.send` using that reference, allowing us to delete `sinks`, `bounce`, and `destination`. That would also avoid retaining any `Workflow.Sink` that is no longer referenced by any `StableSink`

Maintaining `sinks` does add some safety (against [this](https://github.com/square/workflow-swift/blob/4399ddc1c588881b2871603e83b6ab5f4303f219/Workflow/Sources/SubtreeManager.swift#L373-L378), I think?) by ensuring we use the latest `Workflow.Sink` that the `RenderContext` has given us for a given `Action` type. But not total safety, because we're not evicting from `sinks` every sink that wasn't remade during the last `render`.

Is the need for that safety greater when the Rendering holds `StableSink`s rather than functions closing over `Workflow.Sink`s?

PR #258, opened 2023-12-05
[feat]: add 'screen container' API

jamieQ:
## Motivation

in some circumstances it is helpful to know when a type is serving as a logical 'container' for a rendering, and in particular a container of a `Screen`. the types currently exposed by WorkflowUI which serve this purpose are the following:

1. `AnyScreen`
1. `WorkflowHostingController<ScreenType, Output>`
1. `ScreenViewController<ScreenType>`

since both the hosting controller & screen view controller are generic, there is no way to dynamically check if an arbitrary, type-erased type is an instance without also specifying the generic parameters. this makes certain types of dynamic introspection clunky or impossible depending on code structure. to illustrate the point, consider the following somewhat contrived example:

```swift
func makeUninspectableScreenVC() -> UIViewController {
    struct LocalScreen: Screen { ... } // Screen conformance symbol is only visible in this function
    return ScreenViewController(screen: LocalScreen(), environment: .empty)
}

let isSVC = makeUninspectableScreenVC() is ScreenViewController<???> // no generic can be specified here that makes this true
```
we'd like our logical 'Screen container' types to share an interface that allows access to the underlying screen information, which can help with debugging and other runtime data collection.

## Changes

- adds the `ScreenContaining` protocol. this has a single requirement that conformances must provide a `containedScreen` value, returning an `any Screen` type that is the semantically 'contained' screen
- adds conformances to this protocol to the existing screen containers in WorkflowUI
- adds a utility protocol extension to traverse a chain of containers to find the 'leaf' screen

## Checklist

- [x] Unit Tests
- [x] ~UI Tests~
- [x] ~Snapshot Tests (iOS only)~

### DIFF
--- a/WorkflowUI/Sources/Screen/ScreenContaining.swift
+++ b/WorkflowUI/Sources/Screen/ScreenContaining.swift
@@ -1,0 +1,31 @@
+#if canImport(UIKit)
+
+import Foundation
+import UIKit
+
+/// Defines a type that semantically 'contains' a `Screen`, but erases its associated static type information.
+///
+/// This enables referring to `any ScreenViewController` which is not possible without a protocol
+/// like this because `ScreenViewController` is generic over its screen type and it's not possible to check
+/// `viewController is ScreenViewController` without also specifying a specific screen type.
+public protocol ScreenContaining {
+    /// The underlying Screen the conforming type contains
+    var containedScreen: any Screen { get }
+}
+
+extension ScreenContaining {
+    /// Iteratively traverses a sequence of `containedScreen` values until one is found that does _not_
+    /// conform to `ScreenContaining`. Put another way, this returns the first `containedScreen`
+    /// that is a `Screen` and _not_ a `ScreenContaining` type.
+    public func findInnermostScreen() -> any Screen {
+        var result = containedScreen
+
+        while let nextContainer = result as? ScreenContaining {
+            result = nextContainer.containedScreen
+        }
+
+        return result
+    }
+}
+
+#endif
--- a/WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
+++ b/WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
@@ -208,4 +208,12 @@ extension WorkflowHostingController: ViewEnvironmentObserving {
     }
 }
 
+// MARK: ScreenContaining
+
+extension WorkflowHostingController: ScreenContaining {
+    public var containedScreen: any Screen {
+        workflowHost.rendering.value
+    }
+}
+
 #endif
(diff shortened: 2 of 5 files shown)


WorkflowUI/Sources/Screen/ScreenContaining.swift
```diff
@@ -0,0 +1,31 @@
...
+
+import Foundation
+import UIKit
+
+/// Defines a type that semantically 'contains' a `Screen`, but erases its associated static type information.
+///
+/// This enables referring to `any ScreenViewController` which is not possible without a protocol
+/// like this because `ScreenViewController` is generic over its screen type and it's not possible to check
+/// `viewController is ScreenViewController` without also specifying a specific screen type.
+public protocol ScreenContaining {
```
kyleve, line 11: So I see this is still a draft, but how will this work for things like navigation controllers or tab bar controllers, where they have >= 1 inner screens?

kyleve: (I think you may want two properties, `containedScreens : [any Screen]` and `visibleScreens: [any Screen]`?)

amorde: Those only have one current screen so it's ok for those to simply return the current "top" screen for conforming to this protocol.

The harder example is something like a split screen. In that case we may have to opt for tagging that screen with its own identity rather than trying to dig out the two individual screens its composing, but not sure.

amorde (approved):

WorkflowUI/Sources/Screen/ScreenContaining.swift
jamieQ: i think the main goals of this interface are to:

1. add parent type to ScreenViewController<T> so that its generic parameter can be ignored without losing access to its boxed Screen value
2. provide a mechanism for 'traversing' nested containers

i agree there are some questions about particular types of containers that might be worth further consideration, but i'm not sure if we need to deal with those questions at the moment or can punt them if this simple model works for most existing use cases.

mjohnson12 (approved):

watt (approved):

WorkflowUI/Sources/Screen/ScreenContaining.swift
bencochran: Maybe this spelling of the protocol should be `SingleScreenContaining`? That leaves the door open to a `ScreenContaining` that can handle multiple contained screens (where `SingleScreenContaining` could provide its own conformance to _that_ protocol).

I do also somewhat like calling this `visibleContainedScreen` (or similar) to make it clear that, for example, we’re interested in the topmost screen in a navigation stack not all the obscured screens.

jamieQ: thanks for the feedback. i've made an update to rename to the protocol to `SingleScreenContaining`, and its requirement to `primaryScreen` to better communicate the intent. also had a go at improving the associated commentary.

i plan to work on the integration story with the intent to merge this early next week. if you have additional thoughts, do let me know!

PR #265, opened 2024-01-11
[chore]: relax swiftui experimental dependency constraints & bump version

jamieQ:
### Description

the current experimental swift ui target is versioned independently from the other pods. this change relaxes the version requirements for its dependencies to allow any minor or patch changes to be made without this podspec having to be changed.

### DIFF
--- a/WorkflowSwiftUIExperimental.podspec
+++ b/WorkflowSwiftUIExperimental.podspec
@@ -2,7 +2,7 @@ require_relative('version')
 
 Pod::Spec.new do |s|
     s.name         = 'WorkflowSwiftUIExperimental'
-    s.version      = '0.1'
+    s.version      = '0.2'
     s.summary      = 'Infrastructure for Workflow-powered SwiftUI'
     s.homepage     = 'https://www.github.com/square/workflow-swift'
     s.license      = 'Apache License, Version 2.0'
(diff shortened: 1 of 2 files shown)


WorkflowSwiftUIExperimental.podspec
```diff
@@ -18,8 +18,8 @@ Pod::Spec.new do |s|
 
     s.source_files = 'WorkflowSwiftUIExperimental/Sources/*.swift'
 
-    s.dependency 'Workflow', WORKFLOW_VERSION
-    s.dependency 'WorkflowUI', WORKFLOW_VERSION
+    s.dependency 'Workflow', "~> #{WORKFLOW_MAJOR_VERSION}.0"
+    s.dependency 'WorkflowUI', "~> #{WORKFLOW_MAJOR_VERSION}.0"
```
nabs-m, line 22: Say a new major version of Workflow comes out at 4.0.0, seems like that's only when you'd want this dependency to change? So you'd wanna ignore things like 3.6.1, 3.7.0 as well as 4.1.0, 4.1.1, etc? Just wanna make sure I understand the intention here.

nabs-m (approved):
I think I get what you're trying to do, just wanted to confirm with an example.

WorkflowSwiftUIExperimental.podspec
jamieQ: yeah, this change means that version 0.2 of this pod requires version 3.x.y of its two dependencies, but any one will do.

square-tomb, line 22: A possible alternative is hardcoding the version bound like this, which could save you from having to touch this file at all when you are bumping the major version of the other modules. But then you'd still have to bump WorkflowSwiftUIExperimental in a separate change for Register's sake.

Also, I feel like there may have been some other reason I didn't do this 🤔 

```suggestion
    s.dependency 'Workflow', "~> 3.0"
    s.dependency 'WorkflowUI', "~> 3.0"
```

square-tomb (approved):

PR #267, opened 2024-02-06
[chore]: update codeowners

jamieQ:
### Issue

- make more UI-related code have shared ownership with @square/ui-systems-ios

sanity-checked the wildcard logic [here](https://github.com/square/workflow-swift/pull/268)


n8chur (approved):
Thanks!

PR #270, opened 2024-02-13
Removed unnecessary nested generic type

mjohnson12:
Nested type ExpectedSideEffect was using the same name for the workflow generic type as the outer type RenderTester.  This created a shadowed type name warning.
Removing the generic type got rid of the warning

## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowTesting/Sources/Internal/RenderExpectations.swift
+++ b/WorkflowTesting/Sources/Internal/RenderExpectations.swift
@@ -48,7 +48,7 @@ extension RenderTester {
 }
 
 extension RenderTester {
-    internal class ExpectedSideEffect<WorkflowType: Workflow> {
+    internal class ExpectedSideEffect {
         let key: AnyHashable
         let file: StaticString
         let line: UInt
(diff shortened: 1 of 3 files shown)


nabs-m (approved):

WorkflowTesting/Sources/Internal/RenderExpectations.swift
```diff
@@ -48,7 +48,7 @@ extension RenderTester {
 }
 
 extension RenderTester {
-    internal class ExpectedSideEffect<WorkflowType: Workflow> {
+    internal class ExpectedSideEffect<RenderedWorkflowType: Workflow> {
```
jamieQ, line 51: are these generic parameters even necessary? can they differ from the value from the enclosing scope (RenderTester)?

mjohnson12: If I removed them then it failed to compile because ExpectedSideEffect didn't have a generic parameter.

mjohnson12: I was wondering the same thing if they could differ.

mjohnson12: I was able to remove the type and change the declarations where it was used and it seems to work.  I didn't see anywhere where the WorkflowType was different for the side effect vs RenderTester.  Pushing up changes for review.

jamieQ (approved):

PR #271, opened 2024-02-16
!chore: add watchOS support to ViewEnvironment and bump iOS to 15.0

n8chur:
# Overview

Adds watchOS 8 support to `ViewEnvironment` and bumps iOS minimum deployment target to 15.0.

We noticed the need to add watchOS support when we found that [Market publishing was failing](https://mobuild.sqprod.co/blue/organizations/jenkins/CI%2Bmar%2B2RW6NR%2BPublish-iOS-Market-Libraries/detail/CI%2Bmar%2B2RW6NR%2BPublish-iOS-Market-Libraries/3847/pipeline).

# Why

`ViewEnvironment` is imported in `MarketSwiftUI` which requires watchOS support, and all of our upstream dependencies at square are now on the iOS 15 minimum deployment target so it seems like it might be worth bumping just for future convenience when contributing to this workflow-swift.

## Checklist

- [x] Unit Tests
- [x] UI Tests


kyleve (approved):

Package.swift
```diff
@@ -6,8 +6,9 @@ import PackageDescription
 let package = Package(
     name: "Workflow",
     platforms: [
-        .iOS("14.0"),
+        .iOS("15.0"),
```
lickel, line 9: We should be using symbolic references here:

```swift
        .iOS(.v15),
        .macCatalyst(.v15),
        .tvOS(.v15),
        .watchOS(.v8),
        .macOS(.v12),
```

Note: added macOS and catalyst references

watt (approved):

mjohnson12 (approved):

Package.swift
n8chur: Sounds good: de1f0803

n8chur:
I moved to using version number presets in the Package.swift and added tvOS/macCatalyst cc @mjohnson12
de1f0803

mjohnson12 (approved):

PR #277, opened 2024-03-26
[feat]: add support for sizingOptions to SwiftUIScreen

n8chur:
Market's Modal framework relies on `preferredContentSize` to size modals during presentation.

Without any changes, view controllers that back `SwiftUIScreen`s never had their `preferredContentSize` updated. 

`SwiftUIScreen` is backed by a `UIHostingController` subclass. In iOS 16 and above, a new `sizingOption` parameter was added that allows a consumer to specify the `.preferredContentSize` option which will cause the view controller to update the `preferredContentSize` as it changes over time. Before iOS 16, you have to do this manually.

One thing that's surprising is that with the `sizingOption` support for `preferredContentSize`, the `preferredContentSize` does not appear to be updated after the first layout. Also, the `UIHostingController`'s `view` is does not appear to have `setNeedsLayout` set, which has generally been expected in Market's infrastructure so far. I've opened [UI-5797](https://block.atlassian.net/browse/UI-5797) to investigate this to see whether Market's Modal framework is holding this wrong, or if the workarounds found in this PR are in fact needed.

Market also supports having the `preferredContentSize` automatically calculated if it's requested by a consumer via an associated value on `UIViewController`—we'll likely want to figure out a way to support this with `SwiftUIScreen`s, and that may involve moving some of that infra into the Workflow repo ([UI-5801](https://block.atlassian.net/browse/UI-5801)).

I've also introduced a test target for `WorkflowSwiftUIExperimental` so that I could add some tests related to `preferredContentSize` calculations.

An integration experiment was added to Market in [here](https://github.com/squareup/market/pull/8106).

These changes resolve [UI-5766](https://block.atlassian.net/browse/UI-5766).

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
+++ b/WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
@@ -30,7 +30,7 @@ public protocol SwiftUIScreen: Screen {
 }
 
 public extension SwiftUIScreen {
-    static var isEquivalent: ((Self, Self) -> Bool)? { return nil }
+    static var isEquivalent: ((Self, Self) -> Bool)? { nil }
 }
 
 public extension SwiftUIScreen where Self: Equatable {
@@ -87,6 +87,16 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("not implemented")
+    }
+
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        let size = view.sizeThatFits(view.frame.size)
+
+        if preferredContentSize != size {
+            preferredContentSize = size
+        }
     }
 }
 
--- a/WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
+++ b/WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
@@ -26,7 +26,13 @@ public protocol SwiftUIScreen: Screen {
     @ViewBuilder
     static func makeView(model: ObservableValue<Self>) -> Content
 
+    static var sizingOptions: SwiftUIScreenSizingOptions { get }
+
     static var isEquivalent: ((Self, Self) -> Bool)? { get }
+}
+
+public extension SwiftUIScreen {
+    static var sizingOptions: SwiftUIScreenSizingOptions { [] }
 }
 
 public extension SwiftUIScreen {
@@ -46,6 +52,7 @@ public extension SwiftUIScreen {
                 let (model, modelSink) = ObservableValue.makeObservableValue(self, isEquivalent: Self.isEquivalent)
                 let (viewEnvironment, envSink) = ObservableValue.makeObservableValue(environment)
                 return ModeledHostingController(
+                    swiftUIScreenSizingOptions: Self.sizingOptions,
                     modelSink: modelSink,
                     viewEnvironmentSink: envSink,
                     rootView: WithModel(model, content: { model in
@@ -64,6 +71,17 @@ public extension SwiftUIScreen {
     }
 }
 
+public struct SwiftUIScreenSizingOptions: OptionSet {
+    public let rawValue: Int
+
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
+    }
+
+    public static let intrinsicContentSize: SwiftUIScreenSizingOptions = .init(rawValue: 1 << 0)
+    public static let preferredContentSize: SwiftUIScreenSizingOptions = .init(rawValue: 1 << 1)
+}
+
 private struct EnvironmentInjectingView<Content: View>: View {
     @ObservedObject var viewEnvironment: ObservableValue<ViewEnvironment>
     let content: Content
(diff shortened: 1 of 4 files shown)


WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
```diff
@@ -77,17 +95,104 @@ private struct EnvironmentInjectingView<Content: View>: View {
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.backgroundColor = .clear
+
+        setNeedsLayoutBeforeFirstLayoutIfNeeded()
+    }
+
+    override func viewDidLayoutSubviews() {
```
kyleve, line 137: > Market also supports having the preferredContentSize automatically calculated if it's requested by a consumer via an associated value on UIViewController—we'll likely want to figure out a way to support this with SwiftUIScreens, and that may involve moving some of that infra into the Workflow repo ([UI-5801](https://block.atlassian.net/browse/UI-5801)).

Hmm yeah this is an interesting one, since that stuff all lives in Modals right now. We'll definitely want to keep it, since avoiding the duplicative preferredContentSize calculations was a pretty big performance win.

Trying to think if there's a way we could flow this into the SwiftUI environment at the Market layer... But I guess since this is a view controller, we don't have that available.

One gross idea, but it would work, is do a `respondsToSelector` check on the associated value we added, and reference it via `valueForKey` in this repo. Icky but it would do the thing we want...

n8chur: I think we may want to consider moving that associated object to a `ViewEnvironmentKey`, and then move that key's definition to `ViewEnvironmentUI`, possibly also adding support for this automatically in `WorkflowUI` (along with here).

Let's move further discussion on this to [UI-5801](https://block.atlassian.net/browse/UI-5801) so that the context is easier to discover once we start that work!

WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
```diff
@@ -77,17 +95,104 @@ private struct EnvironmentInjectingView<Content: View>: View {
...
     }
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("not implemented")
     }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.backgroundColor = .clear
```
square-tomb, line 132: Just curious: is this new and necessary for the PCS behavior we want?

square-tomb (approved):

WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
n8chur: Oh yeah, whoops, meant to call this out or pull it into a separate PR...

It's not necessary to achieve the goal of this PR, but IMO it's generally expected, and annoying to work around if you want to have a transparent background when working with `Screens` like this. We also set this in `WorkflowHostingController`.

Epoxy sets this on [their hosting controller](https://github.com/airbnb/epoxy-ios/blob/e25f2915a004f64c27ebff7ed9ad41ec594a4cef/Sources/EpoxyCore/SwiftUI/EpoxySwiftUIHostingController.swift#L38-L42) and they included a comment describing their motivation:
```swift
    // A `UIHostingController` has a system background color by default as it's typically used in
    // full-screen use cases. Since we're using this view controller to place SwiftUI views within
    // other view controllers we default the background color to clear so we can see the views
    // below, e.g. to draw highlight states in a `CollectionView`.
```

Happy to move this to a separate PR if we'd prefer!

square-tomb: Thanks for explaining. Makes sense to me, and not important to break out IMO 👍

nsillik: worth adding a similar comment here?

watt (approved):

WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
n8chur: > worth adding a similar comment here?

Eh, maybe? I think it's probably pretty self-explanatory.
Happy to add one if anyone feels strongly though!

n8chur:
Merging now that I have 2 stamps but more than happy to follow up with additional changes if requested!

robmaceachern (approved):

WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
n8chur: > Maybe it's worth cribbing that comment into our implementation too.

2 comments about it seems like a signal that I should add the comment—I'll do so in a follow up!

n8chur: #278

nsillik (commented):
surprisingly straightforward, LGTM!

PR #276, opened 2024-03-15
WorkflowSwiftUI Perception

watt:
Building on @square-tomb's work in #260, this PR prototypes the Perception approach more fully.

This PR uses a fork of CocoaPods to pull in the SPM dependencies. Run `bundle install` and `pod gen` as normal after pulling down this branch and the dependencies should appear in the Xcode project automatically.

additional info TBD

PR #283, opened 2024-05-20
Observable State for WorkflowSwiftUI

watt:
This is a complete implementation of Workflow-powered SwiftUI views using fine-grained observability to minimize render impact. It's based on the prototype in #276, and @square-tomb's previous prototype #260. Squares can learn more at go/workflow-swiftui.

# Observation

We're depending on Perception, Point-Free's backport of Observation. The approach to observable state is adapted from TCA's approach, with a custom `@ObservableState` macro that endows struct types with a concept of identity.

In workflows, you must annotate your state with `@ObservableState`, and render a type conforming to `ObservableModel`, which wraps your state and sends mutations into the unidirectional flow. There are several built-in conveniences for rendering common cases, or you can create a custom type.

On the view side, your `View` will consume a `Store<Model>` wrapper, which provides access to state, sinks, and any child stores from nested workflows.

To wire things up, you can implement a trivial type conforming to `ObservableScreen` and map your rendering. It's *strongly* recommended to keep these layers separate: render a model, implement a view, and create a screen only where needed. This allows for compositions of workflows and views, which is difficult or impossible if rendering a screen directly.

Check out the new ObservableScreen sample app for complete examples of most concepts being introduced here. I'll also write up an adoption guide that expands on each requirement.

# SwiftPM

The `@ObservableState` macro means we cannot ship `WorkflowSwiftUI` with CocoaPods. For now, this PR only removes WorkflowSwiftUI podspec, but it may be preferable to remove all podpsecs and migrate everything to SwiftPM to reduce the maintenance burden.

To work on WorkflowSwiftUI locally, you can use [xcodegen](https://github.com/yonaskolb/XcodeGen) to create a project, similarly to how `pod gen` works.

## Checklist

- [x] Unit Tests
- [x] UI Tests

## Follow-up tasks

- ~~Write an adoption guide~~ (#292)
- Screen integration tests
- Remove podspecs
- Replace xcodegen with Tuist
- Improve the screen-backing UIHostingController (UI-5797)


WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -0,0 +1,208 @@
...
+    /// The sizing options for the screen.
+    var sizingOptions: SwiftUIScreenSizingOptions { get }
+    /// The model that this screen observes.
+    var model: Model { get }
+
+    /// Constructs the root view for this screen. This is only called once to initialize the view.
+    /// After the initial construction, the view will be updated by injecting new values into the
+    /// store.
+    @ViewBuilder
+    static func makeView(store: Store<Model>) -> Content
```
kyleve, line 36: Can/how/should this have access to the Environment?

kyleve: (In general, this being static feels a bit weird to me, I assume it's to reduce access to instance properties, but is that a hard requirement?)

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -0,0 +1,208 @@
...
+
+    public static let preferredContentSize: SwiftUIScreenSizingOptions = .init(rawValue: 1 << 0)
+}
+
+private struct ViewEnvironmentModifier: ViewModifier {
+    @ObservedObject var holder: ViewEnvironmentHolder
+
+    func body(content: Content) -> some View {
+        content
+            .environment(\.viewEnvironment, holder.viewEnvironment)
```
kyleve, line 80: Confirming it's OK to stamp the entire view env modifier over the existing one? No way anything in the middle can change it?

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -0,0 +1,208 @@
...
+        self.viewEnvironment = viewEnvironment
+    }
+}
+
+private final class ModeledHostingController<Model, Content: View>: UIHostingController<ModifiedContent<Content, ViewEnvironmentModifier>> {
+    let setModel: (Model) -> Void
+    let setViewEnvironment: (ViewEnvironment) -> Void
+
+    var swiftUIScreenSizingOptions: SwiftUIScreenSizingOptions {
+        didSet {
```
kyleve, line 97: Should this have a `guard` on not equal?

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -0,0 +1,208 @@
...
+        // to clear to allow this kind of flexibility.
+        view.backgroundColor = .clear
+
+        setNeedsLayoutBeforeFirstLayoutIfNeeded()
+    }
+
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        defer { hasLaidOutOnce = true }
```
kyleve, line 148: I don't think we want this in a `defer` right? Once this method is called, we have laid out once.

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -0,0 +1,208 @@
...
+        if #available(iOS 16.0, *) {
+            // Handled in initializer, but set it on first layout to resolve a bug where the PCS is
+            // not updated appropriately after the first layout.
+            // UI-5797
+            if !hasLaidOutOnce,
+               swiftUIScreenSizingOptions.contains(.preferredContentSize)
+            {
+                let size = view.sizeThatFits(view.frame.size)
+
+                if preferredContentSize != size {
```
kyleve, line 159: Do we want this in `viewDidLayoutSubviews` or `viewWillLayoutSubviews`? We've rightly run into problems in Market doing layout updates too late, in `viewDidLayoutSubviews`, when `viewWillLayoutSubviews` is more correct.

WorkflowSwiftUI/Sources/ObservableState.swift
kyleve, line 1: Is this copied from TCA? If so could we nest all the copied code into a single folder?

WorkflowSwiftUI/Sources/ObservableState.swift
```diff
@@ -0,0 +1,181 @@
...
+/// Conforming to this protocol signals to other APIs that the value type supports observation.
+/// However, applying the ``ObservableState`` protocol by itself to a type doesn’t add observation
+/// functionality to the type. Instead, always use the ``ObservableState()`` macro when adding
+/// observation support to a type.
+#if !os(visionOS)
+public protocol ObservableState: Perceptible {
+    var _$id: ObservableStateID { get }
+    mutating func _$willModify()
+}
+#else
```
kyleve, line 15: Do we want this else at all? Eg doesn't `Perceptible` vs. `Observable` support different types (structs vs classes) for observation?

WorkflowSwiftUI/Sources/ObservableState.swift
```diff
@@ -0,0 +1,181 @@
...
+    mutating func _$willModify()
+}
+#endif
+
+/// A unique identifier for a observed value.
+public struct ObservableStateID: Equatable, Hashable, Sendable {
+    @usableFromInline
+    var location: UUID {
+        get { storage.id.location }
+        set {
```
kyleve, line 27: Should this bail if the new value is the same as the old value?

WorkflowSwiftUI/Sources/ObservableState.swift
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    private var storage: Storage
+
+    private init(storage: Storage) {
+        self.storage = storage
+    }
+
+    public init() {
+        self.init(storage: Storage(id: .location(UUID())))
```
kyleve, line 42: Is the UUID here used for uniquing locations? Could we utilize the #file and #line magic values instead?

WorkflowSwiftUI/Sources/ObservationStateRegistrar.swift
kyleve, line 1: Same Q as above, not sure if this is from TCA or not; might be nice to denote via folder structure

WorkflowSwiftUI/Sources/Store.swift
```diff
@@ -0,0 +1,453 @@
...
+/// For state properties that are writable, an automatic `Binding` can be derived by annotating the
+/// store with `@Bindable`. These bindings will use the workflow's state mutation sink.
+///
+/// All properties can be turned into bindings by appending `sending(store:action:)` or
+/// `sending(closure:)` to specify the "write" action. For properties that are already writable,
+/// this will refine the binding to send a custom action instead of the built-in state mutation
+/// sink.
+///
+@dynamicMemberLookup
+public final class Store<Model: ObservableModel>: Perceptible {
```
kyleve, line 27: Given "Provides access to a workflow's state and actions from within an ``ObservableScreen``", I wonder if this has a clearer name than just `Store`, eg, `ObservableScreenContext` or something.

WorkflowSwiftUI/Sources/Store.swift
```diff
@@ -0,0 +1,453 @@
...
+        self.model = model
+    }
+
+    var state: State {
+        _$observationRegistrar.access(self, keyPath: \.state)
+        return model.accessor.state
+    }
+
+    private func send<Value>(keyPath: WritableKeyPath<State, Value>, value: Value) {
+        guard !invalidated else {
```
kyleve, line 52: If this is false, the change will be silently dropped – should this be a fatal instead, or can we add some sort of additional validation?

WorkflowSwiftUI/Sources/Store.swift
```diff
@@ -0,0 +1,453 @@
...
+        } else {
+            updateModel()
+        }
+
+        // Update and invalidate child stores
+
+        for (keyPath, childStore) in childStores {
+            if childStore.isInvalid(newModel) {
+                childStore.invalidate()
+                childStores[keyPath] = nil
```
kyleve, line 94: Does this have any performance implications? Eg, does this force a redraw of any child screen / nested references?

WorkflowSwiftUI/Sources/Store.swift
```diff
@@ -0,0 +1,453 @@
...
+
+    /// Represents an access to the "wrapper" of a nested child store, such as an Optional or
+    /// collection type.
+    ///
+    /// Each nested model's scope will track its own mutations, but we use this to track mutations
+    /// to the wrapper itself, such as changes to a collection size.
+    internal struct ChildModelAccess {
+        var willSet: (Store<Model>) -> Void
+        var didSet: (Store<Model>) -> Void
+        var changed: (Model, Model) -> Bool
```
kyleve, line 197: Nit: `onChange`?

WorkflowSwiftUI/Sources/Store.swift
```diff
@@ -0,0 +1,453 @@
...
+        isInvalid: @escaping (Model) -> Bool
+    ) -> Store<ChildModel> {
+        if let childStore = childStores[key]?.store as? Store<ChildModel> {
+            return childStore
+        }
+
+        let childModel = getModel(model)
+        let childStore = Store<ChildModel>(childModel)
+
+        childStores[key] = ChildStore(
```
kyleve, line 244: Should this be asserting there isn't already a value stored by this key?

WorkflowSwiftUIMacros/Sources/Availability.swift
kyleve, line 1: Same Q; could we nest files wholly copied from OSS into a folder?

Samples/ObservableScreen/Sources/CounterView.swift
```diff
@@ -0,0 +1,58 @@
...
+import WorkflowSwiftUI
+
+struct CounterView: View {
+    typealias Model = ActionModel<CounterWorkflow.State, CounterWorkflow.Action>
+
+    let store: Store<Model>
+    let key: String
+
+    var body: some View {
+        let _ = Self._printChanges()
```
gilgtm, line 12: Do one of the imports define this function?

Samples/ObservableScreen/Sources/CounterWorkflow.swift
```diff
@@ -0,0 +1,102 @@
...
+            case .increment:
+                state.count += state.info.stepSize
+            case .decrement:
+                state.count -= state.info.stepSize
+            }
+            return nil
+        }
+    }
+
+    typealias Output = Never
```
gilgtm, line 48: Can you add an example feature where this would be used? Or is the idea that it never should be?

RELEASING.md
```diff
@@ -43,7 +43,6 @@ For Squares, membership is managed through the `Workflow Swift Owners` registry
     bundle exec pod trunk push WorkflowRxSwift.podspec --synchronous
     bundle exec pod trunk push WorkflowReactiveSwiftTesting.podspec --synchronous
     bundle exec pod trunk push WorkflowRxSwiftTesting.podspec --synchronous
-    bundle exec pod trunk push WorkflowSwiftUI.podspec --synchronous
```
mjohnson12, line 46: Do we maybe want to update this to explain how you release a new version of WorkflowSwiftUI?
We are currently blocked from doing a Workflow release with Xcode 15 because of an issue in cocoapods.  It would probably benefit us to move all of workflow to package manager if we have to move part of it.

Samples/ObservableScreen/Sources/MultiCounterScreen.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+import Perception
+import SwiftUI
+import ViewEnvironment
+import Workflow
+import WorkflowSwiftUI
+
+struct MultiCounterScreen: ObservableScreen {
```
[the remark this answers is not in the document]
watt: Perhaps, but I wouldn't jump to conclusions from this demo. Most real screens will have a backstack item, for instance.

WorkflowSwiftUI/Sources/ObservableModel.swift
```diff
@@ -0,0 +1,152 @@
...
+///     func render(
+///         state: State,
+///         context: RenderContext<Self>
+///     ) -> StateAccessor<State> {
+///         context.makeStateAccessor(state: state)
+///     }
+/// }
+/// ```
+///
+/// For simple workflows with a single action, you can generate a model from your state and action:
```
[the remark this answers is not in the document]
watt: A single Action type.

WorkflowSwiftUI/Sources/Store.swift
watt: TCA calls it a Store. ¯\_(ツ)_/¯

`ObservableScreenContext` is a mouthful and I'm not sure it really helps you understand the purpose any better than documentation and just learning the new pattern.

WorkflowSwiftUI/Sources/Store.swift
watt: At the top of the function we return the cached store if there's already a store for this key.

WorkflowSwiftUI/Sources/Store.swift
watt: `isChanged` might be more apt — this is called when the model updates to determine if we should register a mutation.

WorkflowSwiftUI/Sources/Store.swift
watt: Nope, releasing a cached child store shouldn't affect views at all. Cached child stores are invalidated if they're no longer accessible, e.g. if a collection changed and the associated element is no longer present. This prevents the cache from growing unbounded.

WorkflowSwiftUI/Sources/Store.swift
watt: Hmm, I'm not sure we can. 🤔

Besides preventing the cache from growing unbounded, invalidating stores is useful because SwiftUI itself has been known to send values into bindings long after their views are invalid, and we don't want those values (and they will potentially assert at the workflow level). I'm not sure there's a way we could tell if a value came from SwiftUI misbehaving or programmer misbehaving.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
```diff
@@ -0,0 +1,102 @@
...
+        self.resetToken = resetToken
+        self.initialValue = initialValue
+        self.maxValue = maxValue
+    }
+
+    func makeInitialState() -> State {
+        State(count: initialValue, maxValue: maxValue, info: info)
+    }
+
+    func workflowDidChange(from previousWorkflow: CounterWorkflow, state: inout State) {
```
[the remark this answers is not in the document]
watt: Yeah, I'll have to cover this in the adoption docs, because it's definitely one of the biggest gotchas.

Parent dependencies need to go on your State so that we can track when they change and avoid body invalidations when they don't change, and `workflowDidChange` is the hook for updating state between two renders.

If a dependency is `@ObservableState` you can assign it directly and the macro handles checking changes. If it's not, you've got to do a comparison yourself. Otherwise, the property will be considered mutated on every render, and invalidate the body of any view that access that property on every render.

I think this is potentially something we could iterate on to make easier in the future, but for now I couldn't come up with anything better.

RELEASING.md
mjohnson12: No I think doing that separately would be good.

Samples/ObservableScreen/Sources/CounterView.swift
```diff
@@ -0,0 +1,58 @@
...
+        WithPerceptionTracking {
```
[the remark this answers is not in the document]
kyleve: Hmm, maybe I'm not 100% understanding, but this more reads to me that this is about leveraging the backport, not that we need this in the view itself... Eg I'm imaging if we had a view that lived right above this that Market owned, that could do something like 

```swift
struct MarketInternalView : View {

   var wrapped : some View

   var body : some View {
      WithPerceptionTracking {
          wrapped.body
       }
   }
}
```

Or similar?

Samples/ObservableScreen/Sources/CounterWorkflow.swift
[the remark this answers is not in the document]
kyleve: I think we should definitely iterate on this! Wonder if a custom macro applied to the workflow itself could deal with this.

Samples/ObservableScreen/Sources/MultiCounterScreen.swift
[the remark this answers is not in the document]
kyleve: True – perhaps something to add to the backlog / optimization list – I just don't love needing to do this

Samples/ObservableScreen/Sources/MultiCounterWorkflow.swift
```diff
@@ -0,0 +1,130 @@
...
+
+        var value: Int
+
+        func apply(toState state: inout MultiCounterWorkflow.State) -> Never? {
+            state.resetToken = .init()
+            return nil
+        }
+    }
+
+    @CasePathable
```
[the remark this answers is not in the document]
kyleve: Makes sense!

Samples/ObservableScreen/Sources/MultiCounterWorkflow.swift
```diff
@@ -0,0 +1,130 @@
...
+        let counters: [CounterWorkflow.Model] = state.counters.map { counter in
+            CounterWorkflow(
+                info: counter,
+                resetToken: state.resetToken,
+                maxValue: maxCounter?.count
+            )
+            .rendered(in: context, key: "\(counter.id)")
+        }
+
+        let sumAction = context.makeSink(of: SumAction.self)
```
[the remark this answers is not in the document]
kyleve: Is this something we expect people to do generally, or this is a special example?

WorkflowSwiftUI/Sources/ObservableScreen.swift
kyleve: Got it, makes sense – curious to see how this is received by folks, since feels a bit weird, but I have no evidence for that!

WorkflowSwiftUI/Sources/ObservableScreen.swift
kyleve: Makes sense!

WorkflowSwiftUI/Sources/ObservableScreen.swift
kyleve: Got it – should it be, though? If I change the sizing param on a screen i'd expect it to apply

WorkflowSwiftUI/Sources/ObservableScreen.swift
kyleve: On the other hand, it might be nice to add more "debt" to that ticket. Thoughts?

WorkflowSwiftUI/Sources/ObservableState.swift
kyleve: I think it'd be nice for current and future us to nest copied-in code in a way that makes it clear it's copied in

WorkflowSwiftUI/Sources/Store.swift
kyleve: Hmm, ok. Curious for Workflow folks' opinions here – I don't love how generic the name is personally

robmaceachern: My vote is to stick with `Store`.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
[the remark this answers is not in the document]
robmaceachern: +1 to spending some time iterating on this. Feels like it's almost certain to cause issues.

robmaceachern (commented):
I'm still digesting Store.swift and others but figured I'd share feedback so far. Will continue review tomorrow!

Samples/ObservableScreen/Sources/MultiCounterView.swift
```diff
@@ -0,0 +1,110 @@
...
+                controls
+
+                if let maxCounter = store.maxCounter {
+                    CounterView(store: maxCounter, key: "max")
+                }
+
+                ForEach(
+                    Array(store.counters.enumerated()),
+                    id: \.element.id
+                ) { index, counter in
```
mjohnson12, line 27: I'm curious when you do the `ForEach` on `counters` which is `CounterWorkflow.Model` what translates that to a `Store<CounterWorkflow.Model>` when you get inside the loop?

mjohnson12: I do really like how you don't need to call to scope the store to a subset of the store when creating a child view.

robmaceachern (approved):
Lots of really cool stuff here. Nice work @watt!

I think it'd be a great idea to schedule a walkthrough for UIS and Foundation folks in the next week or two to improve the bus factor on this stuff.

WorkflowSwiftUI/Tests/StoreTests.swift
```diff
@@ -0,0 +1,821 @@
...
+
+        let countDidChange = expectation(description: "count.didChange")
+
+        withPerceptionTracking {
+            _ = store.count
+        } onChange: {
+            countDidChange.fulfill()
+        }
+
+        var newState = State(count: 1)
```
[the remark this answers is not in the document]
watt: No, the `count` doesn't need to change for the observation to trigger, but later this test is also testing that mutations made through the store affect the model that we pass in. Changing the count just makes it easier to do those assertions.

WorkflowSwiftUI/Tests/ObservableStateTests.swift
```diff
@@ -0,0 +1,297 @@
...
+        let countDidChange = expectation(description: "count.didChange")
+
+        withPerceptionTracking {
+            _ = state.count
+        } onChange: {
+            countDidChange.fulfill()
+        }
+
+        state.count += 1
+        await fulfillment(of: [countDidChange], timeout: 0)
```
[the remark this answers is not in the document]
watt: Basically, yes. No timeout means this call will hang forever if it's not fulfilled, 0 timeout means there's no waiting and the test will fail if the expectations haven't been fulfilled when this is called.

WorkflowSwiftUI/Tests/ObservableStateTests.swift
```diff
@@ -0,0 +1,297 @@
...
+        var state = ParentState()
+        let optionalDidChange = expectation(description: "optional.didChange")
+
+        withPerceptionTracking {
+            _ = state.optional
+        } onChange: {
+            optionalDidChange.fulfill()
+        }
+
+        state.optional = ChildState(count: 42)
```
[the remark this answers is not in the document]
watt: Sure

WorkflowSwiftUI/Sources/ObservableState.swift
watt: OK, I've moved the files that are relatively unchanged copies to "Derived" subfolders.

.github/workflows/swift.yaml
```diff
@@ -61,18 +61,55 @@ jobs:
     - name: Switch Xcode
       run: sudo xcode-select -s /Applications/Xcode_${XCODE_VERSION}.app
 
-    - name: Swift Package Manager - iOS
+    - name: Install XcodeGen
+      run: brew install xcodegen
+
+    - name: Generate Xcode project
+      run: xcodegen generate
```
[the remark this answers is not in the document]
robmaceachern: I'd probably lean towards consolidating to one project generation solution if we can but we can circle back to this later and decide if we want to switch over.

Samples/ObservableScreen/Sources/CounterView.swift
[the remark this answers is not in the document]
kyleve: Got it, makes sense!

I still don't think it's really expected / reasonable to expect everyone to have to wrap like this – perhaps a good case for a macro of our own?

WorkflowSwiftUI/Sources/Store.swift
kyleve: Oh, doh

.github/workflows/swift.yaml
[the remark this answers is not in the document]
kyleve: +1, I think it's best to use one

Samples/ObservableScreen/Sources/MultiCounterWorkflow.swift
```diff
@@ -0,0 +1,125 @@
...
+
+        return MultiCounterModel(
+            accessor: context.makeStateAccessor(state: state),
+            counters: counters,
+            maxCounter: maxCounter,
+            sumAction: sumAction,
+            counterAction: counterAction,
+            resetAction: resetAction
+        )
+    }
```
amorde, line 111: Something that's confusing to me is what this is rendering and how it relates to the views / screen. Sharing the thought process I went through when trying to review this.

It looks like the new expectation is Workflows render a model instead of a screen. But how does that model relate to the eventual screen / view that gets displayed? My first guess was to look at `MultiCounterModel` and see if it has something that produces a screen / view, but it doesn't, so I hit a bit of a dead end because the Workflow doesn't seem to expose any association to the view that gets displayed.

So then I searched around for `MultiCounterView` and found the only place that creates it is the `MultiCounterScreen`. Then I searched for who creates that, and it's actually created by the code that uses `MultiCounterWorkflow`:

```swift
let root = WorkflowHostingController(
    workflow: MultiCounterWorkflow().mapRendering(MultiCounterScreen.init)
)
```

This is jarring because it means that an individual workflow type (in this case `MultiCounterWorkflow`) has no direct tie to the actual screen / view that uses the data from the Workflow. Perhaps this is the intention, but it feels like a weird use of Workflow's paradigms. For example, the name of "rendering" no longer seems to fit here - Workflows don't produce a rendering anymore, they produce a model that an unrelated screen knows how to convert to a SwiftUI view. And if I'm a developer trying to consume a Workflow from another module, how do I know what view I'm supposed to map the rendering to? could we tie them together a bit more so developers don't have to literally read the code to figure out what model connects to which screen?

amorde: Not sure if this would work at all but something that comes to mind is composing the two concepts together, so you have:
* A workflow that renders a model, this implementation of `MultiCounterWorkflow`
* A workflow that renders a screen / view, and consumes the other workflow (in this case `MultiCounterWorkflow`) as a child in it's `render` function and does the translation from `MultiCounterWorkflow.Rendering` to `MultiCounterScreen`

.github/workflows/swift.yaml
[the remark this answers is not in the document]
watt: Alright, I will add a follow-up task to set up a Tuist project for this, assuming that's what we adopt in Market.

Samples/ObservableScreen/Sources/MultiCounterWorkflow.swift
watt: Good feedback! I'll break it up a bit to respond.

> This is jarring because it means that an individual workflow type (in this case `MultiCounterWorkflow`) has no direct tie to the actual screen / view that uses the data from the Workflow. Perhaps this is the intention, but it feels like a weird use of Workflow's paradigms. 

That's correct. The view and the workflow both have a dependency on the state and actions (collectively, the model), but not necessarily on each other. If the state & actions are nested in the workflow type, then the view has a dependency on that, effectively inverting the dependency you'd see if rendering a screen. It's a different pattern, for sure, and it's something we expected to require some education. I'm working on an adoption guide to help with that.

> For example, the name of "rendering" no longer seems to fit here - Workflows don't produce a rendering anymore, they produce a model that an unrelated screen knows how to convert to a SwiftUI view. 

The rendering is, by definition, whatever the workflow renders. The notion that every workflow should be rendering a screen is so pervasive today that I'd say it's harmful — nothing is composable and tests have to force-unwrap type erasers. In a world where views are easily composable, you often won't need a screen to act as the glue, so I'm trying to discourage the idea that every workflow has associated screen type.

> And if I'm a developer trying to consume a Workflow from another module, how do I know what view I'm supposed to map the rendering to? could we tie them together a bit more so developers don't have to literally read the code to figure out what model connects to which screen?

Well in most cases I think it'll be obvious, because they'll live in the same module and have a naming pattern. In other cases, it's a matter of matching up the model. There's no guaranteed 1:1 mapping — you could write 2 different view implementations that take the same model type, for example. You could even power a view with 2 different workflows, but I think that's unlikely.

> Not sure if this would work at all but something that comes to mind is composing the two concepts together, so you have:
> 
> * A workflow that renders a model, this implementation of `MultiCounterWorkflow`
> * A workflow that renders a screen / view, and consumes the other workflow (in this case `MultiCounterWorkflow`) as a child in it's `render` function and does the translation from `MultiCounterWorkflow.Rendering` to `MultiCounterScreen`

You could write a workflow that does that latter step, but that's exactly what `mapRendering()` does!

WorkflowSwiftUI/Sources/ActionModel.swift
```diff
@@ -0,0 +1,24 @@
+/// An ``ObservableModel`` for workflows with a single action.
+///
+/// Rather than creating this model directly, you should use the
+/// ``Workflow/RenderContext/makeActionModel(state:)`` method to create an instance of this model.
+public struct ActionModel<State: ObservableState, Action>: ObservableModel, SingleActionModel {
```
jamieQ, line 5: nit: since `SingleActionModel` refines `ObservableModel` do we need both conformances here? i guess if it doesn't cause any compiler warnings it's fine, and maybe better to be more explicit. curious if there is a functional distinction with the `ObservableModel` conformance removed though

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -0,0 +1,208 @@
...
+/// to the underlying model, and selectively invalidate in response to changes to properties that
+/// are accessed by the view.
+///
+/// Invalidations happen when the observed state is mutated, during actions or the
+/// `workflowDidChange` method. When this screen is rendered, a new model is injected into the
+/// store. Any invalidated views will then be updated with the new model by SwiftUI during its own
+/// rendering cycle.
+///
+/// To use this protocol with a workflow, your workflow should render a type that conforms to
+/// ``ObservableModel``, and then mapped to a screen implementation that uses that concrete model
```
jamieQ, line 19: nit:
```suggestion
/// ``ObservableModel``, and then map to a screen implementation that uses that concrete model
```

watt:
I wrote up an adoption guide in #292.

WorkflowSwiftUI/Sources/ActionModel.swift
watt: No difference. `SingleActionModel` only exists so that it can be used as a constraint in some conveniences for when there's only one action.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    enum Action: WorkflowAction {
+        typealias WorkflowType = CounterWorkflow
+
+        case increment
+        case decrement
+
+        func apply(toState state: inout CounterWorkflow.State) -> CounterWorkflow.Output? {
+            // the max value is applied in the view, not here, so it's possible to increment
```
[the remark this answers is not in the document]
watt: I changed this to apply the max to the state.

Samples/ObservableScreen/Sources/CounterView.swift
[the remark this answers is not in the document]
watt: Can you elaborate on that idea? Not sure how I see how that could be implemented. Open to suggestions and contributions if anyone has a better way!

WorkflowSwiftUI/Tests/ObservableStateTests.swift
```diff
@@ -0,0 +1,297 @@
...
+
+        state.child.reset()
+        await fulfillment(of: [childDidChange], timeout: 0)
+        XCTAssertEqual(state.child.count, 0)
+    }
+
+    func testSwapSiblings() async {
+        var state = ParentState(
+            child: ChildState(count: 1),
+            sibling: ChildState(count: -1)
```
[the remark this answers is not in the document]
watt: fair enough, but since this is a derived file, I'm going leave it as is

WorkflowSwiftUI/Sources/ObservableState.swift
watt: This is a derived file from TCA. Probably the set isn't expensive enough to matter, would be my assumption.

WorkflowSwiftUI/Sources/ObservableState.swift
watt: This is a derived file from TCA. I'm not sure exactly how the location bit works, to be honest.

WorkflowSwiftUI/Sources/ObservationStateRegistrar.swift
watt: Don't recall if I answered this in a previous comment, but all the derivative code has moved to subfolders.

WorkflowSwiftUI/Sources/ObservableScreen.swift
watt: Leaving this for UI-5797 follow-up.

WorkflowSwiftUI/Sources/ObservableModel.swift
```diff
@@ -0,0 +1,152 @@
...
+///     @ObservableState
+///     struct State {
+///         var counter = 0
+///     }
+///
+///     func makeInitialState() -> State {
+///         .init()
+///     }
+///
+///     func render(
```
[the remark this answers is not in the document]
watt: Implements `render` automatically? That would be a highly specific specialization, I think. Probably worth seeing how things are used in some real usages first.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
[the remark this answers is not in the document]
watt: Open to suggestions on this, but I won't have any additional iterations in this PR. Perhaps after we see some real usage, common patterns will be revealed.

Samples/ObservableScreen/Sources/MultiCounterWorkflow.swift
amorde: > You could write a workflow that does that latter step, but that's exactly what mapRendering() does!

Fair! But I was looking for a way for consumers to not have to dig into the code to know what to map the rendering to. Sure they might be similarly named but its still an awkward interface.

> The rendering is, by definition, whatever the workflow renders. The notion that every workflow should be rendering a screen is so pervasive today that I'd say it's harmful — nothing is composable and tests have to force-unwrap type erasers. In a world where views are easily composable, you often won't need a screen to act as the glue, so I'm trying to discourage the idea that every workflow has associated screen type.

100% understand this - I'm more thinking of it as a developer coming the existing norm to this.

I guess the overall feeling is that this doesn't really fit into how we use Workflow today but its still built on top of Workflow as we know it. I wonder if theres some way we could separate the two a bit. What would it look like if you were to implement this from scratch, and how would you want to integrate that into an existing codebase that uses Workflow?

mjohnson12 (approved):

jamieQ (approved):
overall, this seems like an interesting and exciting direction – nice work! as others have pointed out, it is a bit of a break from some existing patterns & paradigms that are fairly prevalent, but that seems somewhat expected given the move to a different UI 'backend' of sorts. some high-level thoughts on things i wish i had a better sense of:

1. how difficult will plugging existing Workflows into this system be? i assume it will depend on the specifics to a large extent, but my impression is that Screen-based renderings won't be easy to re-use (but maybe that's fine & expected).
2. relatedly, what does Worker integration look like in this model? might be good to include that in an example somewhere at some point.
3. [vague thought] what facilities will we suggest to folks for testing the view binding logic? snapshot tests?

i don't really have blocking concerns about the implementation. i think getting some 'real world' feedback on using it would be quite valuable.

WorkflowSwiftUI/Sources/Bindable+Store.swift
```diff
@@ -0,0 +1,124 @@
...
+///     }
+/// }
+///
+/// public struct MyWorkflowView: View {
+///     @Perception.Bindable var store: Store<MyWorkflow.Rendering>
+///
+///     public var body: some View {
+///         Toggle(
+///             "Enabled",
+///             isOn: $store.isOn.sending(action: \.toggle)
```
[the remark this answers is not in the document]
g-mark: +1

WorkflowSwiftUI/Sources/Bindable+Store.swift
```diff
@@ -0,0 +1,124 @@
...
+            keyPath: self.keyPath.appending(path: keyPath)
+        )
+    }
+
+    /// Creates a binding to the value by sending new values through the given sink.
+    ///
+    /// - Parameter sink: The sink to receive an action with values from the binding.
+    /// - Parameter action: An action to contain sent values.
+    /// - Returns: A binding.
+    public func sending<Action>(
```
g-mark, line 96: Just noodling on the API...

- what if the API looked like this from the call site:
   `$store.foo.sending(\.fooAction, to: \.actionSink)`
- Or, given that the func returns a `Binding`, if the func name:
   `$store.foo.binding(\.fooAction, to: \.actionSink)`  
   `$store.foo.bind(to: \.sink, action: \.fooAction)`

Samples/ObservableScreen/Sources/CounterWorkflow.swift
```diff
@@ -0,0 +1,119 @@
...
+    let maxValue: Int?
+
+    @ObservableState
+    struct State {
+        private var _count: Int
+        
+        var count: Int {
+            get { _count }
+            set {
+                if let maxValue, newValue > maxValue {
```
g-mark, line 20: Is there a concern that the logic for protecting state can be located in a setter as well as in an action? I mean, this could probably have been done prior to SwiftUI, but this state is mutable from the SwiftUI view because it's `@ObservableState`, so will be much easier to do.

g-mark (commented):
This is great work, @watt!

Please forgive the no0b questions.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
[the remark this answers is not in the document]
square-tomb: I agree that a macro on the workflow type seems like a promising way reduce boilerplate here, but is a deep enough problem that it should be left for a future iteration.

This is a great example case for that problem, where the workflow's props are a mix of `ObservableState` and non-OS types.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
```diff
@@ -0,0 +1,119 @@
...
+
+        // maxValue is not observable. We should conditionally update it only on change.
+        // Otherwise every set will trigger invalidation.
+        if maxValue != previousWorkflow.maxValue {
+            state.maxValue = maxValue
+        }
+    }
+
+    typealias Rendering = ActionModel<State, Action>
+    typealias Model = ActionModel<State, Action>
```
square-tomb, line 95: This typealias is not necessary to conform to some protocol; it just looks nicer to write `CounterWorkflow.Model` than `CounterWorkflow.Rendering` in some places?

Samples/ObservableScreen/Sources/MultiCounterView.swift
```diff
@@ -0,0 +1,108 @@
...
+                    }
+                    .padding(.vertical, 4)
+                }
+
+                // When showSum is false, changes to counters do not invalidate this body
+                if store.showSum {
+                    HStack {
+                        Text("Sum")
+                        Spacer()
+                        Text("\(store.counters.map(\.count).reduce(0, +))")
```
square-tomb, line 45: This is interesting. WWDC counsels against doing potentially expensive calculations in view bodies (e.g. _O(n)_ string manipulations), but I'm not sure where else you could conveniently locate this kind of calculation across multiple observable models' values.

square-tomb (approved):

Samples/ObservableScreen/Sources/CounterView.swift
```diff
@@ -0,0 +1,61 @@
...
+                    store.send(.decrement)
+                } label: {
+                    Image(systemName: "minus")
+                }
+
+                Text("\(store.count)")
+                    .monospacedDigit()
+
+                Button {
+                    store.send(.increment)
```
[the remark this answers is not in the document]
g-mark: Yes, when I did that locally it prevented both direct mutation and the creation of bindings.

Samples/ObservableScreen/Sources/CounterView.swift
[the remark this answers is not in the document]
watt: > We do have a runtime warning that you'll see if you access an `ObservableState` type's property outside of a `WithPerceptionTracking`, correct?

That's right. You'll get a "purple" runtime warning.

Screenshot 2024-08-08 at 4 29 59 PM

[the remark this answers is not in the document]
kyleve: > and we should expect that [views that take escaping closures](https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/observationbackport/#Lazy-view-closures) will force us to even write multiple WithPerceptionTracking wrappers in the same view body, like

Huh, I thought that Apple had fixed this somehow, but I don't remember how...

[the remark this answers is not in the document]
kyleve: > Can you elaborate on that idea?

I was thinking we could potentially add a type-level macro that could mutate the `body` and add the `WithPerceptionTracking` call, but perhaps that not possible.

kyleve (approved):

Samples/ObservableScreen/Sources/MultiCounterWorkflow.swift
watt: > I guess the overall feeling is that this doesn't really fit into how we use Workflow today but its still built on top of Workflow as we know it. I wonder if theres some way we could separate the two a bit. What would it look like if you were to implement this from scratch, and how would you want to integrate that into an existing codebase that uses Workflow?

Well, you're not wrong, but I think that was sort of the goal, to build a SwiftUI integration on top of Workflow. I'm not sure how we could separate the two further without breaking that. We did prototype some alternative designs that were much more similar to how Workflow is used today, but ultimately opted for this instead because its advantages seemed to be worth the cost of the pattern disruption.

watt:
> 1. how difficult will plugging existing Workflows into this system be? i assume it will depend on the specifics to a large extent, but my impression is that Screen-based renderings won't be easy to re-use (but maybe that's fine & expected).

Highly dependent on the workflow and what your goals for migration are, but you are correct that screens are generally not reusable. If you are replacing a Blueprint screen rendering with a SwiftUI screen, most of the work may go towards rewriting your Element as a View. If you are hoping to use the same workflow to power both a Blueprint screen and SwiftUI screen during the transition, you'll also need to adapt your Blueprint screen to consume/wrap the view model.

I'll probably write a guide specifically for migrations like this after we've seen what that looks like in practice.

> 2. relatedly, what does Worker integration look like in this model? might be good to include that in an example somewhere at some point.

I'm not anticipating any changes there. Workers won't interact directly with SwiftUI, and don't need to observe state.

> 3. [vague thought] what facilities will we suggest to folks for testing the view binding logic? snapshot tests?

Snapshot tests, primarily. XCUI tests may also be very helpful, to fill KIF's shoes, and to validate that state changes only invalidate the views we expect. I have a follow-up task to figure out what that looks like.

Samples/ObservableScreen/Sources/CounterView.swift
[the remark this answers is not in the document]
watt: > Is there any guidance around how to choose between actions, bindings (e.g. for a TextField), or direct mutation? Asked another way, how tightly do we want to embrace/enforce the unidirectional data flow aspect of Workflow?

The unidirectional flow is enforced either way. Mutations from setters like `store.count += 1` are routed through a `StateMutationSink`, which sends a generic `AnyWorkflowAction<WorkflowType>` under the hood.

The choice of when to use a custom action is probably going to be driven by:
- need to do something more than updating the property (maybe you touch 2 properties, maybe there's logging, etc)
- migrating a workflow that already a custom action

In other words, "whenever you need it or want it", heh.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
watt: Yeah. I tried to be consistent with the use of `Model` as a first-class concept.

Samples/ObservableScreen/Sources/CounterWorkflow.swift
watt: Well, it's not exactly the setter that's the concern, but because accessible properties get a "free" mutation action, there is definitely a gotcha where you might declare an action and forget to use it. It'll be easily discoverable if your action does anything, but might still be an annoying lesson to learn.

Samples/ObservableScreen/Sources/MultiCounterView.swift
watt: Hmm, yeah, I'm not sure. If it's expensive it's not great in the workflow render either.

WorkflowSwiftUI/Sources/Bindable+Store.swift
watt: That's not bad! I probably prefer `sending` because `$store.foo` is already a binding, and we're only refining one side of it. I'll keep this in mind for API refinements since we'll likely have others come up too.

g-mark (approved):

PR #285, opened 2024-06-18
[fix]: short circuit worker logging & require explicit opt-in

jamieQ:
### Issue

the existing os logging implementations log various type descriptions as part of the log message. these strings are apparently calculated even if the logging has been disabled, which introduces unneeded/undesirable work in release builds. these logs shouldn't be made in release builds, and should be opt-in in other contexts.

### Description

this change adds short circuiting logic for all the os logging methods in addition to a new environment variable check that can be used to enable the logging when debugging.

- [x] TODO: WorkflowLogger should probably get the same treatment

## Checklist

- [x] Unit Tests – didn't add unit tests, but manually ran existing ones using the new env variable to confirm expected behavior (os_log calls aren't made when the env var isn't set)
~- [x] UI Tests~ n/a
~- [x] Snapshot Tests (iOS only)~ n/a
~- [x] I have made corresponding changes to the documentation~ n/a

### DIFF
--- a/WorkflowCombine/Sources/Logger.swift
+++ b/WorkflowCombine/Sources/Logger.swift
@@ -15,6 +15,7 @@
  */
 
 import os.signpost
+@_spi(Logging) import Workflow
 
 private extension OSLog {
     static let worker = OSLog(subsystem: "com.squareup.WorkflowCombine", category: "Worker")
@@ -29,6 +30,8 @@ final class WorkerLogger<WorkerType: Worker> {
     // MARK: - Workers
 
     func logStarted() {
+        guard WorkflowLogging.isOSLoggingAllowed else { return }
+
         os_signpost(
             .begin,
             log: .worker,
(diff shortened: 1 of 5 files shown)


WorkflowCombine/Sources/Logger.swift
```diff
@@ -14,8 +14,14 @@
  * limitations under the License.
  */
 
+import Foundation
 import os.signpost
 
+private let workerLoggingEnabled: Bool = {
+    let env = ProcessInfo.processInfo.environment
+    return env["com.squareup.workflow.workerLogging"] != nil
```
kyleve, line 22: Should this check for a bool, not just nil? Eg what if this is set to false?

jamieQ: that would probably be a better check. i guess `(... as NSString).boolValue` is perhaps the way to go?

kyleve: I believe that's the standard way to check, yeah

natecraft1 (approved):

natecraft1:
Thanks for jumping on this! I made this ticket https://block.atlassian.net/browse/MPR-2752 so MPR can start to track the hangs that are being addressed

kyleve (approved):

mjohnson12 (approved):

PR #292, opened 2024-08-03
WorkflowSwiftUI adoption guide

watt:
A documentation follow-up to #283.

### DIFF
--- a/Documentation/WorkflowSwiftUI Adoption Guide.md
+++ b/Documentation/WorkflowSwiftUI Adoption Guide.md
@@ -1,0 +1,448 @@
+# WorkflowSwiftUI Adoption Guide
+
+`WorkflowSwiftUI` is designed with performance in mind. Instead of updating the entire UI on every render loop, `WorkflowSwiftUI` uses the [Perception](https://github.com/pointfreeco/swift-perception) framework (a backport of Apple's Observation), to detect the data that each view is dependent on, and to only re-evaluate those views when that data has changed.
+
+In order to support this, `WorkflowSwiftUI` has some additional restrictions and subtle departures from practices you may be used to.
+
+- Your *view* and *workflow* both have knowledge of *state* and *actions* (collectively, a “model”).
+- Your workflow renders a *model* rather than a *screen*.
+- We recommend defining your model, view, and screen independently, rather than nesting any of these types.
+
+## State
+
+Observation is used to detect when the data that views use has changed, and to re-evaluate only the views that are affected.
+
+To achieve this, your workflow’s `State` type must be annotated with the `@ObservableState` macro. This macro adds observation hooks to each property, and also endows your state struct with a concept of identity. 
+
+```swift
+@ObservableState
+public struct State {
+  var foo: String
+  var bar: Int
+}
+```
+
+For an `ObservableState` type, creating a copy does not change the state’s identity in the copy. However, creating a new value by calling the initializer will result in a new identity.
+
+If your state has nested types, you can annotate them with `@ObservableState` as well, for fine-grained observation of the properties of that type as well.
+
+```swift
+@ObservableState
+public struct State {
+  var detail: Detail
+}
+
+@ObservableState
+public struct Detail {
+  var name: String
+  var age: Int
+}
+```
+
+In this example, changes to `state.detail.name` do not affect views that only use `state.detail.age` , and vice-versa. Without annotating `Detail` as `@ObservableState`, any change to a property on `state.detail` will invalidate views using any other property.
+
+## Rendering models
+
+Your workflow should render a type conforming to `ObservableModel`. This provides your view with a way to read from state and send updates that follow Workflow’s unidirectional data flow.
+
+```swift
+public protocol ObservableModel<State> {
+    /// The associated state type that this model observes.
+    associatedtype State: ObservableState
+
+    /// The accessor that can be used to read and write state.
+    var accessor: StateAccessor<State> { get }
+}
+```
+
+Some conveniences are available on the `RenderContext` to make it easy to create a model type.
+
+For trivial workflows with no actions, you can generate a model directly from your state:
+
+```swift
+struct TrivialWorkflow: Workflow {
+    typealias Output = Never
+
+    @ObservableState
+    struct State {
+        var counter = 0
+    }
+
+    func makeInitialState() -> State {
+        .init()
+    }
+
+    func render(
+        state: State,
+        context: RenderContext<Self>
+    ) -> StateAccessor<State> {
+        context.makeStateAccessor(state: state)
+    }
+}
+
+```
+
+For simple workflows with a single action, you can generate a model from your state and action:
+
+```swift
+struct SingleActionWorkflow: Workflow {
+    typealias Output = Never
+
+    @ObservableState
+    struct State {
+        var counter = 0
+    }
+
+    enum Action: WorkflowAction {
+        typealias WorkflowType = SingleActionWorkflow
+        case increment
+
+        func apply(toState state: inout State) -> Never? {
+            state.counter += 1
+            return nil
+        }
+    }
+
+    func makeInitialState() -> State {
+        .init()
+    }
+
+    func render(
+        state: State,
+        context: RenderContext<Self>
+    ) -> ActionModel<State, Action> {
+        context.makeActionModel(state: state)
+    }
+}
+
+```
+
+For complex workflows that have multiple actions or compose observable models from child
+workflows, you can create a custom model that conforms to `ObservableModel`:
+
+```swift
+struct ComplexWorkflow: Workflow {
+    typealias Output = Never
+
+    @ObservableState
+    struct State {
+        var counter = 0
+    }
+
+    enum UpAction: WorkflowAction {
+        typealias WorkflowType = ComplexWorkflow
+        case increment
+
+        func apply(toState state: inout State) -> Never? {
+            state.counter += 1
+            return nil
+        }
+    }
+
+    enum DownAction: WorkflowAction {
+        typealias WorkflowType = ComplexWorkflow
+        case decrement
+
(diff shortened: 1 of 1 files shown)


Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+      // binding with a custom setter action
+      Toggle("Y?", isOn: $store.isY.sending(action: \.setY))
+    }
+  }
+}
+```
+
+## Parent dependencies
+
+Because `WorkflowSwiftUI` relies on observation to function, most properties on your model cannot be safely accessed directly through the `Store` — only the state, child models, and action sinks are accessible. So, when a workflow has dependencies provided by a parent (inputs), they must be added to your workflow’s own state in order to be visible by your view. They must also be updated in the `workflowDidChange` function of your workflow in order to receive updates from your parent.
```
jamieQ, line 383: minor: might be worth also referencing the 'props' terminology for the 'inputs' that is used in some of the existing documentation (maybe mostly on the Kotlin-side).

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+        state: State,
+        context: RenderContext<Self>
+    ) -> ActionModel<State, Action> {
+        context.makeActionModel(state: state)
+    }
+}
+
+```
+
+For complex workflows that have multiple actions or compose observable models from child
```
kyleve, line 120: > For complex workflows that have multiple actions

I feel like this is basically _the_ common case, it'd be nice to make this require less boilerplate/code to be written.

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+}
+```
+
+For an `ObservableState` type, creating a copy does not change the state’s identity in the copy. However, creating a new value by calling the initializer will result in a new identity.
+
+If your state has nested types, you can annotate them with `@ObservableState` as well, for fine-grained observation of the properties of that type as well.
+
+```swift
+@ObservableState
+public struct State {
```
kyleve, line 31: Remind me, but does Apple's observation stuff support `structs` yet, or is it just classes?

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+  var age: Int
+}
+
+enum PersonAction {
+  case giveHighFive
+}
+
+typealias PersonModel = ActionModel<PersonState, PersonAction>
+
+struct PersonView: View {
```
kyleve, line 197: Spitballing on how to avoid people needing `WithPerceptionTracking`:

We define a `ObservedView` protocol,

```swift
protocol ObservingView: View {

   var observing : some View { get }

}

extension ObservingView {
   var body : some View {
      WithPerceptionTracking { observing }
   }
}
```

Where folks would implement `observing`

Which requires learning the type, but once you know, you won't need the perception tracking wrapper, I think?

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+        Button("✋") {
+          store.send(.giveHighFive)
+        }
+      }
+    }
+  }
+}
+```
+
+If you forget to wrap your view's body in `WithPerceptionTracking`, a runtime warning will remind you.
```
kyleve, line 215: Should we turn this into an assert as well? I worry folks will miss the runtime warning. Not sure if this is possible though since I assume this comes from the point free library!

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+    associatedtype Model: ObservableModel
+
+    /// The model that this screen observes.
+    var model: Model { get }
+
+    /// Constructs the root view for this screen. This is only called once to initialize the view.
+    /// After the initial construction, the view will be updated by injecting new values into the
+    /// store.
+    @ViewBuilder
+    static func makeView(store: Store<Model>) -> Content
```
kyleve, line 237: Nit / thought: `view(with:)` or `body(with:)`?

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+}
+```
+
+Your screen will be relatively simple, as it has the relatively simple job of acting as the glue between your rendering model and your view type.
+
+```swift
+struct PersonScreen: ObservableScreen {
```
kyleve, line 244: I think I mentioned this, but it'd be neat if we had a generic type that folks could use instead of needing to define this on their own

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+  let model: PersonModel
+
+  static func makeView(store: Store<PersonModel>) -> PersonView {
+    PersonView(store: store)
+  }
+}
+```
+
+To use this screen, render your workflow and then use the `mapRendering()` function to wrap your workflow’s model rendering in a screen.
```
kyleve, line 253: Nit: include an example?

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+    }
+  }
+}
+```
+
+## Parent dependencies
+
+Because `WorkflowSwiftUI` relies on observation to function, most properties on your model cannot be safely accessed directly through the `Store` — only the state, child models, and action sinks are accessible. So, when a workflow has dependencies provided by a parent (inputs), they must be added to your workflow’s own state in order to be visible by your view. They must also be updated in the `workflowDidChange` function of your workflow in order to receive updates from your parent.
+
+When updating dependencies, remember that any value being set is considered a mutation by the observation framework. To avoid invalidating your view on every render, check if dependencies have actually changed before copying them.
```
kyleve, line 385: Could we add some sort of operator to make this simpler, or a func to help with it? Eg, add a `~=` that checks the values before assigning them?

kyleve (approved):
This is great, thanks! Posting some comments that I think I may've posted on the main PR, but my thoughts are a bit more well-formed now, so repeating myself a bit

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+
+    func makeInitialState() -> State {
+        .init()
+    }
+
+    func render(
+        state: State,
+        context: RenderContext<Self>
+    ) -> ActionModel<State, Action> {
+        context.makeActionModel(state: state)
```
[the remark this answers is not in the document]
robmaceachern: I _think_ the idea is that this used for a single WorkflowAction _type_ (multiple cases is fine). It'd probably be worth adding an additional action case to this example to help illustrate/reinforce that though (and a small tweak to the wording above the example).

Documentation/WorkflowSwiftUI Adoption Guide.md
robmaceachern: The good news is that their runtime warning will fail tests when they are encountered in that context.

Documentation/WorkflowSwiftUI Adoption Guide.md
robmaceachern: @watt Clients will be able to drop `WithPerceptionTracking` once the deployment target is bumped to 17, right?

If that's the case, I probably wouldn't introduce a new wrapper protocol _just_ for that. Maybe we'll end up wanting/needing to wrap up some other functionality/convenience at the Market level and we can figure out how best to slip it in there.

robmaceachern (approved):

Documentation/WorkflowSwiftUI Adoption Guide.md
watt: sure, done

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -0,0 +1,448 @@
...
+@ObservableState
+public struct State {
+  var foo: String
+  var bar: Int
+}
+```
+
+For an `ObservableState` type, creating a copy does not change the state’s identity in the copy. However, creating a new value by calling the initializer will result in a new identity.
+
+If your state has nested types, you can annotate them with `@ObservableState` as well, for fine-grained observation of the properties of that type as well.
```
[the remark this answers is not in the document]
watt: Yes, it's not inherited. I've updated the example to demo that better.

Documentation/WorkflowSwiftUI Adoption Guide.md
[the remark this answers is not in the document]
watt: Updated the example to have another case demonstrating this.

Documentation/WorkflowSwiftUI Adoption Guide.md
watt: Yep, just classes, that's why we need our own `ObservableState` macro.

Documentation/WorkflowSwiftUI Adoption Guide.md
watt: > @watt Clients will be able to drop `WithPerceptionTracking` once the deployment target is bumped to 17, right?

That's right. You also need to use it inline in some lazy contexts, like `GeometryReader`.

Documentation/WorkflowSwiftUI Adoption Guide.md
watt: Added

watt:
I think I've addressed all the documentation-related comments. Punting on the stuff related to API and patterns — I think we covered most of that in the main PR.

Issue #293, opened 2024-08-07
[bug]: RxSwift bindings should observe actions on the MainScheduler


jamieQ:
we have an invariant that Sink emissions must occur on the main thread. this line should probably be changed to `observe(on:)` to prevent inadvertently breaking this expectation.

PR #294, opened 2024-08-07
[fix]: update RxSwift bindings to observe on main thread

jamieQ:
Resolves: #293

## Checklist

- [x] Unit Tests

### DIFF
--- a/WorkflowRxSwift/Sources/ObservableWorkflow.swift
+++ b/WorkflowRxSwift/Sources/ObservableWorkflow.swift
@@ -20,7 +20,7 @@ import class Workflow.Lifetime
 
 extension Observable: AnyWorkflowConvertible {
     public func asAnyWorkflow() -> AnyWorkflow<Void, Element> {
-        return ObservableWorkflow(observable: self).asAnyWorkflow()
+        ObservableWorkflow(observable: self).asAnyWorkflow()
     }
 }
 
@@ -41,6 +41,7 @@ struct ObservableWorkflow<Value>: Workflow {
             let disposable = observable
                 .map { AnyWorkflowAction(sendingOutput: $0) }
                 .subscribe(on: MainScheduler.asyncInstance)
+                .observe(on: MainScheduler.asyncInstance)
                 .subscribe(onNext: { value in
                     sink.send(value)
                 })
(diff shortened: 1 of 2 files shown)


WorkflowRxSwift/Sources/ObservableWorkflow.swift
```diff
@@ -40,7 +40,7 @@ struct ObservableWorkflow<Value>: Workflow {
         context.runSideEffect(key: "") { [observable] lifetime in
             let disposable = observable
                 .map { AnyWorkflowAction(sendingOutput: $0) }
-                .subscribe(on: MainScheduler.asyncInstance)
+                .observe(on: MainScheduler.asyncInstance)
```
jamieQ, line 43: we could additionally leave the `subscribe` logic here... seems like it may not be necessary though if the 'subscription side effects' occur synchronously, since we'll be on the main thread when `subscribe(onNext:)` is executed.

jamieQ: edit: decided to 'play it safe' and just leave the existing code too. not sure what may or may not depend on it.

loganmoseley:
Just to link to documentation: [RxSwift/Documentation/Schedulers.md](https://github.com/ReactiveX/RxSwift/blob/main/Documentation/Schedulers.md) ([permalink](https://github.com/ReactiveX/RxSwift/blob/6491a16da8b7dfa5d33e317b7c6af889a3b297da/Documentation/Schedulers.md)).

WorkflowRxSwift/Sources/ObservableWorkflow.swift
mjohnson12: I'd vote for removing `.subscribe(on:` since it's just wrong but either way `observe(on:` is what we want here.

mjohnson12 (approved):

WorkflowRxSwift/Sources/ObservableWorkflow.swift
jamieQ: yeah... maybe it is. my only argument against that is if removing it changes something that was async to sync, causing other issues. although, i suppose you can make the case that this also changes behavior in a similar way... we're taking things that used to synchronously emit in whatever context they were in to ones that are (always?) queued async on the main thread.

square-tomb (approved):

PR #297, opened 2024-08-22
[fix] View environment customizations on hosting VCs now propagate to SwiftUI environment

robmaceachern:
- View environment customizations on hosting VCs now propagate to SwiftUI environment
- ModeledHostingController implementations now conform to ViewEnvironmentObserving
- Added tests

These changes ensure that any view environment customizations made to a Swift UI screen's backing VC are propagated to the SwiftUI environment. We hit this issue in Market when attempting to propagate the VC's navigation item to SwiftUI and there may be other uses for this behaviour too.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowSwiftUI/Tests/ObservableScreenTests.swift
+++ b/WorkflowSwiftUI/Tests/ObservableScreenTests.swift
@@ -1,0 +1,80 @@
+#if canImport(UIKit)
+
+import SwiftUI
+import ViewEnvironment
+@_spi(ViewEnvironmentWiring) import ViewEnvironmentUI
+import XCTest
+@testable import WorkflowSwiftUI
+
+final class ObservableScreenTests: XCTestCase {
+    func test_viewEnvironmentObservation() {
+        // Ensure that environment customizations made on the view controller
+        // are propagated to the SwiftUI view environment.
+
+        var state = MyState()
+
+        let viewController = TestKeyEmittingScreen(
+            model: MyModel(
+                accessor: StateAccessor(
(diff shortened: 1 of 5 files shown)


WorkflowSwiftUI/Tests/ObservableScreenTests.swift
```diff
@@ -0,0 +1,80 @@
+#if canImport(UIKit)
+
+import SwiftUI
+import ViewEnvironment
+@_spi(ViewEnvironmentWiring) import ViewEnvironmentUI
+import XCTest
+@testable import WorkflowSwiftUI
```
robmaceachern, line 7: @watt The `@testable` was required here because the `StateAccessor` init is not public. Not sure if that was intentional or not. I can open it up as part of this if you'd like!

mjohnson12 (approved):

WorkflowSwiftUI/Tests/ObservableScreenTests.swift
watt: hmm, probably OK if it's public.

watt (approved):

WorkflowSwiftUI/Tests/ObservableScreenTests.swift
robmaceachern: 👍🏽 https://github.com/square/workflow-swift/pull/297/commits/46961b0924ae027cd40aa1a8197fc86abab9fe56

PR #299, opened 2024-08-28
[release]: bump to version 3.11.0

jamieQ:


version.rb
```diff
@@ -2,7 +2,7 @@
 
 # The primary version numbers for Workflow-related pods
 WORKFLOW_MAJOR_VERSION ||= '3'
-WORKFLOW_MINOR_VERSION ||= '10'
```
jamieQ, line 5: cc @watt @robmaceachern – there wasn't anything in the recent commits that warrants anything more than a minor version bump, was there?

robmaceachern: Minor bump sounds reasonable to me! The biggest set of changes was related to [Observable State for WorkflowSwiftUI](https://github.com/square/workflow-swift/commit/b027b1d8ee43095ad133c56d347910560bd886ce) but I think that was all additive.

robmaceachern (approved):

mjohnson12 (approved):

version.rb
watt: Yeah I would consider that a minor bump. There was a deprecation but no removals.

PR #304, opened 2024-10-02
fix: explicit test dependencies

watt:
These weren't being picked up on CI but could fail to build locally. The Tuist option to enforce dependencies should help detect similar stuff in the future.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -173,7 +173,7 @@ let package = Package(
         ),
         .testTarget(
             name: "WorkflowRxSwiftTests",
-            dependencies: ["WorkflowRxSwiftTesting"],
+            dependencies: ["WorkflowRxSwiftTesting", "WorkflowReactiveSwift"],
             path: "WorkflowRxSwift/Tests"
         ),
         .target(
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -60,7 +60,7 @@ let package = Package(
         .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", from: "7.1.1"),
         .package(url: "https://github.com/ReactiveX/RxSwift.git", from: "6.6.0"),
         .package(url: "https://github.com/nicklockwood/SwiftFormat", exact: "0.54.0"),
-        .package(url: "https://github.com/swiftlang/swift-syntax", from: "509.0.0"),
+        .package(url: "https://github.com/swiftlang/swift-syntax", "509.0.0"..<"601.0.0-prerelease"),
```
watt, line 63: Not related to the broken deps, but something I wanted to update while I was in here.

mjohnson12 (approved):
Thanks for the fix!

PR #306, opened 2024-10-03
Changed WorkflowConcurrency Worker task priority to high

mjohnson12:
## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowConcurrency/Sources/Worker.swift
+++ b/WorkflowConcurrency/Sources/Worker.swift
@@ -30,6 +30,7 @@ public protocol Worker<Output>: AnyWorkflowConvertible where Rendering == Void {
     associatedtype Output
 
     /// Execute the work represented by this worker asynchronously and return the result.
+    /// This function will be run in a Task set to high priority.
     func run() async -> Output
     /// Returns `true` if the other worker should be considered equivalent to `self`. Equivalence should take into
     /// account whatever data is meaningful to the task. For example, a worker that loads a user account from a server
@@ -63,7 +64,7 @@ struct WorkerWorkflow<WorkerType: Worker>: Workflow {
         let sink = context.makeOutputSink()
         context.runSideEffect(key: state) { lifetime in
             let send: @MainActor(Output) -> Void = sink.send
-            let task = Task {
+            let task = Task(priority: .high) {
                 logger.logStarted()
                 let output = await worker.run()
                 if Task.isCancelled {


nabs-m (approved):

WorkflowConcurrency/Sources/Worker.swift
```diff
@@ -29,6 +29,9 @@ public protocol Worker<Output>: AnyWorkflowConvertible where Rendering == Void {
     /// The type of output events returned by this worker.
     associatedtype Output
 
+    /// Task priority to use to run the async function.
+    var priority: TaskPriority? { get }
```
jamieQ, line 33: i haven't totally thought this through, but my reflexive reaction is to not want to add this property if we don't have to... is there some way we can just carry a value from the callsite where the actual work is defined? also, if we add this it sort of implies more state associated with the worker – how is that handled for the equivalence check?

mjohnson12: The reason for the property was to allow consumers who are defining their own workers to also be able to set a task priority.
Which equivalence check? I can look into that.

mjohnson12: Also currently the workflow running `AsyncOperationWorker` just has access to it as a `Worker` protocol so not sure how it would get access to the priority otherwise.  We could create a new workflow that is typed `AsyncOperationWorker` instead but that seems overkill if we also want `Worker`s to be able to set a priority.

jamieQ: hmm yeah i think i kinda see... what do you think of just internally setting the Task to high priority? pros: leaves public API alone, is simple, and is probably the 'right' default (since Workflow is effectively a UI-bound framework at the moment). cons: implicit behavior that someone might not want? if a client _really_ wants to get something to run at a lower priority for some reason, they could explicitly detach a task with the desired priority.

mjohnson12: I guess since this is a one async function call and done worker I could live with that.  I do think it would be cool to essentially support the same options as `Task` in `AsyncOperationWorker` but if it causes problems with equality just using high priority might be our best compromise.

nabs-m (approved):
I think it's fair to assume UI related tasks are always gonna be of high priority

WorkflowConcurrency/Sources/Worker.swift
```diff
@@ -63,7 +63,7 @@ struct WorkerWorkflow<WorkerType: Worker>: Workflow {
         let sink = context.makeOutputSink()
         context.runSideEffect(key: state) { lifetime in
             let send: @MainActor(Output) -> Void = sink.send
-            let task = Task {
+            let task = Task(priority: .high) {
```
jamieQ, line 66: maybe add a note in one of the doc comments about this behavior

jamieQ (approved):

PR #309, opened 2024-10-09
Improve releasing instructions

nabs-m:
Updating instructions for 2 main reasons:
1. Be explicit about what's needed for consumption via SPM.
2. Add all the steps actually required to publish to CocoaPods to remove guesswork and debugging.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,37 +2,40 @@
 
 ## Production Releases
 
----
-
-***Before you begin:*** Please make sure you are set up with 
-[`bundle exec pod trunk`](https://guides.cocoapods.org/making/getting-setup-with-trunk.html) and your CocoaPods
-account is a contributor for all pods. If you need to be added as a
-contributor, please [open a ticket requesting access](https://github.com/square/workflow-swift/issues/new).
-
-For Squares, membership is managed through the `Workflow Swift Owners` registry group. Please request access to that group through Registry. Once you have access, you can register a session with `bundle exec pod trunk` using the group e-mail alias `workflow-swift-owners@squareup.com`.
-
-`bundle exec pod trunk register workflow-swift-owners@squareup.com 'Workflow Swift Owners' --description='Your computer description'`
-
----
-
-> ⚠️ [Optional] To avoid possible headaches when publishing podspecs, validation can be performed before updating the Workflow version number(s). To do this, run the following in the root directory of this repo:
-> ```bash
-> bundle exec pod lib lint Workflow.podspec ViewEnvironment.podspec ViewEnvironmentUI.podspec WorkflowTesting.podspec WorkflowReactiveSwift.podspec WorkflowUI.podspec WorkflowRxSwift.podspec WorkflowReactiveSwiftTesting.podspec WorkflowRxSwiftTesting.podspec WorkflowSwiftUIExperimental.podspec WorkflowCombine.podspec WorkflowCombineTesting.podspec WorkflowConcurrency.podspec WorkflowConcurrencyTesting.podspec
-> ```
-> You may need to `--include-podspecs` for pods that have changed and are depended on by other of the pods.
-
+### Updating the release
 1. Update `WORKFLOW_VERSION` in `version.rb` based on [`semver`](https://semver.org/).
 
 1. Create a PR with the version change and merge to `main`.
 
 1. Create the release on GitHub:
    1. Go to the [Releases](https://github.com/square/workflow-swift/releases) and `Draft a new release`.
-   1. `Choose a tag` and create a tag for the new version. ex: `v1.0.0`
-   1. `Generate release notes`
+   1. `Choose a tag` and create a tag for the new version. ex: `v1.0.0`.
+   1. `Generate release notes`.
    1. Ensure the `Title` corresponds to the version we're publishing and the generated `Release Notes` are accurate.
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -2,37 +2,48 @@
...
+At this point, the new release is available and can be consumed using the Swift Package Manager.
+
+[Square specific] To make the new version available internally, [bump the version](https://go/spm-bump-dependency) through SPM.
+
+### Publishing to CocoaPods
+
+1. You need to have 2 older Xcode versions, but you will only use them through the command line, so do not worry about the app being incompatible with your macOS version. You will also need to tweak one of them:
+   1. Download **Xcode 14.2**. This will be used to publish all pods except for `WorkflowRxSwift` and `WorkflowRxSwiftTesting`.
+   1. Download **Xcode 15.4**. This will be used to publish `WorkflowRxSwift` and `WorkflowRxSwiftTesting`.
+   1. Copy `Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc`from within the Xcode 14.2 app to the same location in Xcode 15.4.
```
jamieQ, line 26: is it clear why this workaround is needed? the workaround description seems accurate to me, but obviously isn't ideal. curious if there's something we can/should change instead so we don't have to do the 2 versions of Xcode dance.

RELEASING.md
```diff
@@ -2,37 +2,48 @@
...
+1. To avoid possible headaches when publishing podspecs, validation can be performed first.
+    1. Switch to Xcode 15.4: `sudo xcode-select -s Path/to/Xcode15.4`.
+    1. Run `bundle exec pod lib lint WorkflowRxSwift.podspec WorkflowRxSwiftTesting.podspec`.
+    1. Ensure you're on Xcode 14.2: `sudo xcode-select -s Path/to/Xcode14.2`.
+    1. Run the following:
+    ```bash
+    bundle exec pod lib lint Workflow.podspec ViewEnvironment.podspec ViewEnvironmentUI.podspec WorkflowTesting.podspec WorkflowReactiveSwift.podspec WorkflowUI.podspec WorkflowReactiveSwiftTesting.podspec WorkflowSwiftUIExperimental.podspec WorkflowCombine.podspec WorkflowCombineTesting.podspec WorkflowConcurrency.podspec WorkflowConcurrencyTesting.podspec
+    ```
+
+1. Once validation passes on all pods, you can publish to CocoaPods:
```
jamieQ, line 46: i thought the Xcode dance had to be done during the actual `pod trunk push` calls for the RxSwift dependencies

jamieQ (approved):
proposed updates seem accurate to me! i do wonder if we should just try to fix the weird issue with the `/usr/lib/arc` thing instead of having to document the workarounds, but also realize we're hoping to migrate to just SPM reasonably soon, so that might not be worthwhile.

RELEASING.md
```diff
@@ -2,37 +2,48 @@
...
-   1. `Generate release notes`
+   1. `Choose a tag` and create a tag for the new version. ex: `v1.0.0`.
+   1. `Generate release notes`.
    1. Ensure the `Title` corresponds to the version we're publishing and the generated `Release Notes` are accurate.
    1. Hit `Publish release`.
 
-1. Publish to CocoaPods:
+At this point, the new release is available and can be consumed using the Swift Package Manager.
+
+[Square specific] To make the new version available internally, [bump the version](https://go/spm-bump-dependency) through SPM.
```
mjohnson12, line 19: Should we put anything Square specific in a public .md file?

RELEASING.md
nabs-m: It has to do with a CocoaPods issue complaining about the macOS deployment target being too low. More details [here](https://github.com/CocoaPods/CocoaPods/issues/12033).
Like you said, it's not worthwhile digging too deep with us potentially moving away from CocoaPods and the workaround isn't too costly (copying that directory only need to happen once, and everything else seems to behave fine with it).

RELEASING.md
nabs-m: That's right, I missed clarifying that this part also needs to be done with the 2 Xcode versions

RELEASING.md
nabs-m: I think go links that has the actual details is fine, [this line](https://github.com/square/workflow-swift/pull/309/files#diff-87d6a57a759f2386bf9caf5309b9aa0b9ff0ca70f5e38696e114a1ccdb0e785dR64) about the same thing with CocoaPods was already there.

RELEASING.md
mjohnson12: when I did this my experience was the opposite for the versions.  Xcode 15 did not work for RxSwift and 14.2 did.

nabs-m: > when I did this my experience was the opposite for the versions. Xcode 15 did not work for RxSwift and 14.2 did.

Interesting. I did have issues with Xcode 15 but now I'm wondering if that was before I copied that directory, and only tried it with the `RxSwift` pods. I'll try again in case all we need is Xcode 15 + the `arc` folder.

nabs-m: Look at that, Xcode 15.4 can indeed do them all now 🎉 Updated the docs to remove the version dance, but we still have to copy the `arc` directory.

mjohnson12 (approved):

PR #311, opened 2024-11-14
feat: expose helpers for static models

watt:
Creating a static model is useful for testing, and the missing `public init` on `ActionModel` makes that difficult to do. This PR adds that init along with a couple static factory methods for that use case.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowSwiftUI/Sources/ActionModel.swift
+++ b/WorkflowSwiftUI/Sources/ActionModel.swift
@@ -1,10 +1,22 @@
 /// An ``ObservableModel`` for workflows with a single action.
 ///
-/// Rather than creating this model directly, you should use the
-/// ``Workflow/RenderContext/makeActionModel(state:)`` method to create an instance of this model.
+/// To create an accessor, use
+/// ``Workflow/RenderContext/makeActionModel(state:)``. State writes and actions
+/// will be sent to the workflow.
 public struct ActionModel<State: ObservableState, Action>: ObservableModel, SingleActionModel {
     public let accessor: StateAccessor<State>
     public let sendAction: (Action) -> Void
+
+    /// Creates a new ActionModel.
+    ///
+    /// Rather than creating this model directly, you should usually use the
+    /// ``Workflow/RenderContext/makeActionModel(state:)`` method to create an
+    /// instance of this model. If you need a static model for testing or
+    /// previews, you can use the ``constant(state:)`` method.
+    public init(accessor: StateAccessor<State>, sendAction: @escaping (Action) -> Void) {
+        self.accessor = accessor
+        self.sendAction = sendAction
+    }
 }
 
 /// An observable model with a single action.
@@ -22,3 +34,15 @@ extension ActionModel: Identifiable where State: Identifiable {
         accessor.id
     }
 }
+
+#if DEBUG
+
+public extension ActionModel {
+    /// Creates a static model which ignores all sent values, suitable for static previews
+    /// or testing.
+    static func constant(state: State) -> ActionModel<State, Action> {
+        ActionModel(accessor: .constant(state: state), sendAction: { _ in })
+    }
+}
+
+#endif
--- a/WorkflowSwiftUI/Sources/Store+Preview.swift
+++ b/WorkflowSwiftUI/Sources/Store+Preview.swift
@@ -12,23 +12,17 @@ public struct StaticStorePreviewContext {
     }
 
     public func makeStateAccessor<State>(state: State) -> StateAccessor<State> {
-        StateAccessor(
-            state: state,
-            sendValue: { _ in }
-        )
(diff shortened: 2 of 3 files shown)


WorkflowSwiftUI/Sources/Store+Preview.swift
```diff
@@ -12,23 +12,17 @@ public struct StaticStorePreviewContext {
...
-        ActionModel(
-            accessor: makeStateAccessor(state: state),
-            sendAction: makeSink(of: Action.self).send
-        )
+        .static(state: state)
     }
 }
 
-extension Store {
+public extension Store {
```
g-mark, line 25: Please keep the access control on the individual methods.

g-mark (commented):
These are nice additions!

Two comments:

- If I could bike-shed for a moment, `static` feels close, but also ambiguous due to the overloaded meaning in Swift. Hoping there might be a more explicit name. Some ideas: noOp, noOperation, identity, ignored, ignore, ignoringChanges/ignoringActions, immutable.

- Noticed this warning (Worker.swift, line 77):

  > passing argument of non-sendable type 'WorkerType.Output' into main actor-isolated context may introduce data races

  We should mark `Output` as `Sendable` soon.

WorkflowSwiftUI/Sources/ActionModel.swift
```diff
@@ -5,6 +5,11 @@
 public struct ActionModel<State: ObservableState, Action>: ObservableModel, SingleActionModel {
     public let accessor: StateAccessor<State>
     public let sendAction: (Action) -> Void
+
+    public init(accessor: StateAccessor<State>, sendAction: @escaping (Action) -> Void) {
```
jamieQ, line 9: are there any reasons we'd want to restrict clients' ability to create these types in general?

WorkflowSwiftUI/Sources/ActionModel.swift
```diff
@@ -22,3 +27,15 @@ extension ActionModel: Identifiable where State: Identifiable {
         accessor.id
     }
 }
+
+#if DEBUG
+
+public extension ActionModel {
+    /// Creates a static model which ignores all sent values, suitable for static previews
+    /// or testing.
+    static func `static`(state: State) -> ActionModel<State, Action> {
```
jamieQ, line 36: per @g-mark's naming thoughts – this form of utility could perhaps be named '`just`' if we want to lean into Combine's parlance

jamieQ (approved):

WorkflowSwiftUI/Sources/ActionModel.swift
robmaceachern: Or follow [`Binding.constant(_:)`](https://developer.apple.com/documentation/swiftui/binding/constant(_:))'s lead.

robmaceachern (approved):

WorkflowSwiftUI/Sources/Store+Preview.swift
watt: I agree, but the linter did this!

jamieQ: approve me! https://github.com/square/workflow-swift/pull/315

WorkflowSwiftUI/Sources/ActionModel.swift
watt: Sure

WorkflowSwiftUI/Sources/ActionModel.swift
watt: To reduce the chance of folks using it wrong, mainly. You can't create something like a `RenderContext` either. The type's docs already hint at this, and I can copy it to the init as well.

watt:
> * If I could bike-shed for a moment, `static` feels close, but also ambiguous due to the overloaded meaning in Swift. Hoping there might be a more explicit name. Some ideas: noOp, noOperation, identity, ignored, ignore, ignoringChanges/ignoringActions, immutable.

Changed to `constant` per Rob's suggestion.

> * Noticed this warning (Worker.swift, line 77):
>   > passing argument of non-sendable type 'WorkerType.Output' into main actor-isolated context may introduce data races
>   
>   We should mark `Output` as `Sendable` soon.

Feel free to take that on! I think it might take some integration work since it'll affect a lot of types.

WorkflowSwiftUI/Sources/ActionModel.swift
jamieQ: to be clear – you think the convenience of exposing the public constructor outweighs the chance of misuse?

g-mark:
> >   We should mark `Output` as `Sendable` soon.
> 
> Feel free to take that on! I think it might take some integration work since it'll affect a lot of types.

On it!

PR #313, opened 2024-11-14
Make Worker.Output Sendable

g-mark:
Adds `Sendable` conformance as a constraint on `Workflow.Output`.

## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowCombine/Sources/Worker.swift
+++ b/WorkflowCombine/Sources/Worker.swift
@@ -28,7 +28,7 @@ import Workflow
 /// If there is, and if the workers are 'equivalent', the context leaves the existing worker running.
 ///
 /// If there is not an existing worker of this type, the context will kick off the new worker (via `run`).
-public protocol Worker<Output>: AnyWorkflowConvertible where Rendering == Void {
+public protocol Worker<Output>: AnyWorkflowConvertible where Rendering == Void, Output: Sendable {
     /// The type of output events returned by this worker.
     associatedtype Output
     associatedtype WorkerPublisher: Publisher where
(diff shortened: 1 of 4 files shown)


mjohnson12 (commented):
Will this break register?  I would be surprised if all Workers Outputs were Sendable.

g-mark:
Integration passed 👍

jamieQ (commented):
before we proceed further with this effort, i think we should have a bit of discussion around the motivation and future direction of this type of change to better expose things to the concurrency system. two thoughts that come to mind:

1. imposing a `Sendable` conformance on a type is a fairly substantial ask in general since it's asserting that type must be able to be passed across arbitrary isolation domains. is this the right approach?
2. is this potentially a source-break for existing clients, so would necessitate a major semver bump? (granted register is probably the primary one)

g-mark:
> before we proceed further with this effort, i think we should have a bit of discussion around the motivation and future direction of this type of change to better expose things to the concurrency system.

That makes sense. For background, this came up because of a [warning](https://github.com/square/workflow-swift/pull/311/files#file-workflowconcurrency-sources-worker-swift-L77) currently in `WorkflowConcurrency` while calling `await send(output)`.

> 1. imposing a `Sendable` conformance on a type is a fairly substantial ask in general since it's asserting that type must be able to be passed across arbitrary isolation domains. is this the right approach?

This is _probably_ the right approach, given what `Sendable` types are. Although I struggle to think of another approach, I would love to hear all other ideas.

> 2. is this potentially a source-break for existing clients, so would necessitate a major semver bump? (granted register is probably the primary one)

Yes it _does_ have the potential to break clients if their `Output` types do not conform to `Sendable`, so I agree a major bump would be appropriate (side note: Alamofire changed a bunch of things to `Sendable` in a minor bump, and it broke some of our code).

PR #314, opened 2024-11-14
Remove empty workflowDidChange from Xcode project templates

blakemcanally:
Remove the empty workflowDidChange(from previousWorkflow:state:) method from Xcode project templates and tests.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Samples/AsyncWorker/Sources/AsyncWorkerWorkflow.swift
+++ b/Samples/AsyncWorker/Sources/AsyncWorkerWorkflow.swift
@@ -24,8 +24,6 @@ extension AsyncWorkerWorkflow {
     func makeInitialState() -> AsyncWorkerWorkflow.State {
         State(model: Model(message: "Initial State"))
     }
-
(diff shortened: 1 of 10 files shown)


jamieQ (approved):
while you're in here, would you mind seeing if we can remove [this one](https://github.com/square/workflow-swift/blob/a29d9d1753119a38421bfff1a862c324eb14da92/Workflow/Sources/Workflow.swift#L95) as well? seems like it is redundant since we have a blanket default implementation (though i haven't personally tried compiling it).

blakemcanally:
> while you're in here, would you mind seeing if we can remove [this one](https://github.com/square/workflow-swift/blob/a29d9d1753119a38421bfff1a862c324eb14da92/Workflow/Sources/Workflow.swift#L95) as well? seems like it is redundant since we have a blanket default implementation (though i haven't personally tried compiling it).

updated! lets see how CI likes it

PR #315, opened 2024-11-14
[gardening]: swiftformat – prefer acl's on declarations

jamieQ:
changing the lint preference for this repo from having ACLs on extensions to declarations. this just changes the config – i did not run it against the repo yet.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -36,6 +36,7 @@
 --empty void
 --exponentcase lowercase
 --exponentgrouping disabled
+--extensionacl on-declarations
 --fractiongrouping disabled
 --fragment false
 --hexgrouping none


watt:
Worth syncing up with Register? Or do we want a unique set of rules here?

jamieQ:
> Worth syncing up with Register? Or do we want a unique set of rules here?

personally i prefer this rule, so would be in favor of adopting universally. since we manage things independently here though i think it's fine to proceed with this without blocking on anyone's having stronger opinions about a larger change. i can float the proposed change to see what people think.

watt:
> > Worth syncing up with Register? Or do we want a unique set of rules here?
> 
> personally i prefer this rule, so would be in favor of adopting universally. since we manage things independently here though i think it's fine to proceed with this without blocking on anyone's having stronger opinions about a larger change. i can float the proposed change to see what people think.

It's already present there, just I'm questioning whether we should be maintaining a separate config file in general.

g-mark (commented):
_Thank You_

jamieQ:
> It's already present there, just I'm questioning whether we should be maintaining a separate config file in general.

jamieQ:
> It's already present there, just I'm questioning whether we should be maintaining a separate config file in general.

ah i see, thanks for elaborating. yeah having bifurcation here doesn't make sense to me – i'll copy over the config as best it can be.

jamieQ:
actually it looks to me like this specific rule is _not_ yet enabled in register, so will plan to merge this before looking into a broader synchronization effort

g-mark (approved):

PR #319, opened 2025-01-28
Display a deprecation warning and fatal error when the top-level `@Environment(\.viewEnvironment)` is referenced

kyleve:
Display a deprecation warning and fatal error when the top-level `@Environment(\.viewEnvironment)` is referenced, to enforce people reference a sub-property. This ensures proper SwiftUI view invalidation.

image

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
+++ b/ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
@@ -29,4 +29,28 @@ extension EnvironmentValues {
     }
 }
 
+
+extension Environment where Value == ViewEnvironment {
+    
+    @available(
+        *,
+         deprecated,
+         message:
+            """
+            Please do not create an `@Environment` property that references the top-level `viewEnvironment`: \
+            it will break SwiftUI's automatic invalidation when any part of the `ViewEnvironment` changes. \
+            Instead, reference your relevant sub-property, eg `@Environment(\\.viewEnvironment.myProperty)`.
+            """
+    )
+    @inlinable public init(_ keyPath: KeyPath<EnvironmentValues, Value>) {
+        fatalError(
+            """
+            Please do not create an `@Environment` property that references the top-level `viewEnvironment`: \
+            it will break SwiftUI's automatic invalidation when any part of the `ViewEnvironment` changes. \
+            Instead, reference your relevant sub-property, eg `@Environment(\\.viewEnvironment.myProperty)`.
+            """
+        )
+    }
+}
+
 #endif


ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
```diff
@@ -29,4 +29,28 @@ extension EnvironmentValues {
...
+        *,
+         deprecated,
+         message:
+            """
+            Please do not create an `@Environment` property that references the top-level `viewEnvironment`: \
+            it will break SwiftUI's automatic invalidation when any part of the `ViewEnvironment` changes. \
+            Instead, reference your relevant sub-property, eg `@Environment(\\.viewEnvironment.myProperty)`.
+            """
+    )
+    @inlinable public init(_ keyPath: KeyPath<EnvironmentValues, Value>) {
```
kyleve, line 45: I can't think of a case where we'd want to do this, but if we want to allow it, perhaps adding an SPI'd `init(unsafe: ...)` would allow that escape hatch.

robmaceachern (approved):
Nice! Thanks for doing this.

mjohnson12 (approved):
Nice work

ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
```diff
@@ -29,4 +29,28 @@ extension EnvironmentValues {
...
+        fatalError(
```
jamieQ, line 46: is fatal'ing unconditionally the right approach? what exactly is the downside if one does this? if it's 'just' poor performance, should we be more lenient with the production code and make this an assertion?

ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
```diff
@@ -29,4 +29,28 @@ extension EnvironmentValues {
...
+
+extension Environment where Value == ViewEnvironment {
+    
+    @available(
+        *,
+         deprecated,
+         message:
+            """
+            Please do not create an `@Environment` property that references the top-level `viewEnvironment`: \
+            it will break SwiftUI's automatic invalidation when any part of the `ViewEnvironment` changes. \
```
jamieQ, line 41: for my own personal edification: can you elaborate on the risk & consequences a bit more? what is the behavioral difference in these 2 approaches?

jamieQ (approved):
looks reasonable to me

ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
kyleve: I think a fatal is correct / preferred, since you should never do this in production code. If we already had consumers of SwiftUI doing this, I'd probably go w/ an assert, but given we have the opportunity to "close the door" on this permanently, I'd say let's do that

johnnewman-square (approved):

ViewEnvironment/Sources/EnvironmentValues+ViewEnvironment.swift
kyleve: So basically, the way that SwiftUI does state invalidation is to compare the properties on a view between the old view and the new view, and if they're different or otherwise not comparable, the view is redrawn and updated.

Because `ViewEnvironment` is a big opaque bag o' crap, its possible that there's non-comparable stuff in there, or there's stuff in there that changes every render cycle, meaning that it'd lead to view invalidation. So the fix for that is to scope your `Environment` property down to just the data you need.

I _know_ there's a point free or objc.io article about this in more depth somewhere, however I am struggling to find it.

jamieQ: thanks! this makes sense – i had failed to realize that the root `.viewEnvironment` key is like putting a copy of _another_ environment into the... `SwiftUI.Environment`, and clearly you don't want to depend on the entire environment itself, just the stuff your view 'cares about'.

PR #320, opened 2025-01-28
feat: flexible SwiftUI preferredContentSize calc [UI-7665]

watt:
This is a possible solution for preferredContentSize of SwiftUI content, for discussion.

A summary of the problem:
- The system provided PCS calculation is unconstrained and not useful to us. We need to calculate this manually.
- SwiftUI ScrollViews greedily fill the space available, but will report the natural content size when given an infinite size.

This solution finds an appropriate PCS for the general case by measuring in a few different constraints and combining the results. This accommodates content with no scroll view, as well as content with scroll views in either axis or both.

~~If we like this I'll follow up with tests that validate PCS in various permutations of size and scrolling axes.~~ tests added!

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowSwiftUI/Sources/ObservableScreen.swift
+++ b/WorkflowSwiftUI/Sources/ObservableScreen.swift
@@ -133,10 +133,9 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        // `UIHostingController`'s provides a system background color by default. In order to
-        // support `SwiftUIScreen`s being composed in contexts where it is composed within another
-        // view controller where a transparent background is more desirable, we set the background
-        // to clear to allow this kind of flexibility.
+        // `UIHostingController` provides a system background color by default. We set the
+        // background to clear to support contexts where it is composed within another view
+        // controller.
         view.backgroundColor = .clear
 
         setNeedsLayoutBeforeFirstLayoutIfNeeded()
--- a/WorkflowSwiftUI/Sources/ObservableScreen.swift
+++ b/WorkflowSwiftUI/Sources/ObservableScreen.swift
@@ -97,14 +97,15 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
     var swiftUIScreenSizingOptions: SwiftUIScreenSizingOptions {
         didSet {
             updateSizingOptionsIfNeeded()
-
-            if !hasLaidOutOnce {
+            if isViewLoaded {
                 setNeedsLayoutBeforeFirstLayoutIfNeeded()
             }
         }
     }
 
     private var hasLaidOutOnce = false
+    private var maxFrameWidth: CGFloat = 0
+    private var maxFrameHeight: CGFloat = 0
 
     init(
         setModel: @escaping (Model) -> Void,
@@ -133,7 +134,7 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
         super.viewDidLoad()
 
         // `UIHostingController`'s provides a system background color by default. In order to
-        // support `ObervableModelScreen`s being composed in contexts where it is composed within another
+        // support `SwiftUIScreen`s being composed in contexts where it is composed within another
         // view controller where a transparent background is more desirable, we set the background
         // to clear to allow this kind of flexibility.
         view.backgroundColor = .clear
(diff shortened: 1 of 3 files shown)


WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -146,21 +147,31 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
...
+
+            let fixedSize = CGSize(width: width, height: height)
+
+            // Measure a few different ways to account for ScrollView behavior. ScrollViews will
+            // always greedily fill the space available, but will report the natural content size
+            // when given an infinite size. By combining the results of these measurements we can
+            // deduce the natural size of content that scrolls in either direction, or both, or
+            // neither.
+
+            let fixedResult = view.sizeThatFits(fixedSize)
```
kyleve, line 167: Any idea what the perf cost is for these repeated measurements? I wonder if we could introduce our own sizing options that would allow us to do fewer calls based on more specific sizing configurations.

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -146,21 +147,31 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
...
+            let unboundedHorizontalResult = view.sizeThatFits(CGSize(width: .infinity, height: fixedSize.height))
+            let unboundedVerticalResult = view.sizeThatFits(CGSize(width: fixedSize.width, height: .infinity))
+
+            let size = CGSize(
+                width: min(fixedResult.width, unboundedHorizontalResult.width),
+                height: min(fixedResult.height, unboundedVerticalResult.height)
+            )
```
jamieQ, line 174: any chance we need to (or should) handle weird/unexpected values here? is `sizeThatFits` guaranteed to return sensical results? i guess the concern would be if we end up assigning a preferred content size that doesn't 'make sense' – would that matter?

WorkflowSwiftUIExperimental/Sources/SwiftUIScreen.swift
```diff
@@ -144,21 +145,31 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
...
-            {
-                let size = view.sizeThatFits(view.frame.size)
-
-                if preferredContentSize != size {
-                    preferredContentSize = size
-                }
-            }
-        } else if swiftUIScreenSizingOptions.contains(.preferredContentSize) {
-            let size = view.sizeThatFits(view.frame.size)
+        if swiftUIScreenSizingOptions.contains(.preferredContentSize) {
```
jamieQ, line 148: if this is the path we want to go, shall we extract the duplicate calculation logic?

WorkflowSwiftUI/Sources/ObservableScreen.swift
jamieQ: also, is the new logic specifically for scroll views? if so, do we want to do some of these extra calculations conditionally? or does it have to account for them being embedded as descendant views?

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -146,21 +147,31 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
 
         defer { hasLaidOutOnce = true }
 
-        if #available(iOS 16.0, *) {
```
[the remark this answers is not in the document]
watt: Yep. Previously we were letting `UIHostingController` handle PCS on iOS 16+, except for the first layout. This `if` handled the first layout case, and the `else` was for ≤15. Now we're handling PCS in all cases.

WorkflowSwiftUI/Sources/ObservableScreen.swift
watt: > Any idea what the perf cost is for these repeated measurements

Unknown, but my suspicion is it's fine due to SwiftUI layout being much better than what we're used to.

> also, is the new logic specifically for scroll views? if so, do we want to do some of these extra calculations conditionally? or does it have to account for them being embedded as descendant views?

We do need to calculate PCS regardless. The multiple measurements are to account for scroll views. Unfortunately we have to account for there being a scroll view anywhere in the content, and there's just no way to know.

WorkflowSwiftUI/Sources/ObservableScreen.swift
watt: > any chance we need to (or should) handle weird/unexpected values here? is `sizeThatFits` guaranteed to return sensical results? i guess the concern would be if we end up assigning a preferred content size that doesn't 'make sense' – would that matter?

It's not guaranteed to do anything by API contract, but based on observation it seems like SwiftUI will return sensible values.

johnnewman-square (approved):
After trying out these changes in the Market Catalog app, SwiftUI Screens size themselves much better than before 🎉. I also didn't experience performance hiccups on a physical device with dialogs and sheets. I think this is a good solution to merge in!

jamieQ (approved):
changes all look reasonable to me!

mjohnson12 (approved):

WorkflowSwiftUI/Tests/PreferredContentSizeTests.swift
```diff
@@ -0,0 +1,123 @@
...
+        WithPerceptionTracking {
+            if store.axes.isEmpty {
+                box
+            } else {
+                ScrollView(store.axes) {
+                    box
+                }
+            }
+        }
+    }
```
robmaceachern, line 116: I'd love if multiline text was also included in the test cases.

WorkflowSwiftUI/Sources/ObservableScreen.swift
robmaceachern: > Unfortunately we have to account for there being a scroll view anywhere in the content, and there's just no way to know.

It'd be a science experiment, but I wonder how feasible it'd be to push the measurements onto the SwiftUI side with a custom Layout and then bubble up context like 'hasScrollView'/'hasMultilineText' via preferences so we could be slightly smarter with our measurements.

I think I'd rather spend the time submitting feedback to Apple to hopefully make this better directly out of the box with the hosting controller though!

robmaceachern (approved):
Thanks for working through this!

WorkflowSwiftUI/Sources/ObservableScreen.swift
```diff
@@ -133,7 +134,7 @@ private final class ModeledHostingController<Model, Content: View>: UIHostingCon
         super.viewDidLoad()
 
         // `UIHostingController`'s provides a system background color by default. In order to
-        // support `ObervableModelScreen`s being composed in contexts where it is composed within another
+        // support `SwiftUIScreen`s being composed in contexts where it is composed within another
```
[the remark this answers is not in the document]
watt: Whoops. I feel like this line is self-explanatory but [apparently not](https://github.com/square/workflow-swift/pull/277#discussion_r1541943323).

WorkflowSwiftUI/Tests/PreferredContentSizeTests.swift
watt: Would be hard to do without making the test fragile.

PR #323, opened 2025-03-06
Rename `expect(worker:producingOutput:)` to be more clear about what it is asserting

amayers:
When I first saw tests that used this function, I thought it was asserting that the `worker` passed in, matches the `workflow.worker` (which it does do), and that it was asserting that `producingOutput == workflow.worker.output` (which it does NOT do). Instead, neither of the workers are ever run. The `producingOutput` is used as mock output. So renaming these to make it more clear what is going on.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/WorkflowCombine/Testing/WorkerTesting.swift
+++ b/WorkflowCombine/Testing/WorkerTesting.swift
@@ -21,15 +21,35 @@ import XCTest
 @testable import WorkflowCombine
 
 extension RenderTester {
-    /// Expect the given worker. It will be checked for `isEquivalent(to:)` with the requested worker.
+    /// Mock the given worker's output, and assert that the workflow's worker `isEquivalent(to:)` the `worker`.
     ///
     /// - Parameters:
-    ///   - worker: The worker to be expected
-    ///   - producingOutput: An output that will be returned when this worker is requested, if any.
+    ///   - worker: The worker that we expect was created by the workflow. Will be compared using
+    ///             `isEquivalent(to:)` to assert that the workflow's worker matches.
+    ///   - producingOutput: The output to be used instead of actually running the worker.
+    ///                      If the workflow never tries to run the worker, then this won't be used.
     ///   - key: Key to expect this `Workflow` to be rendered with.
+    @available(*, deprecated, renamed: "expectedWorker(_:mockingOutput:key:file:line:)", message: "renamed")
     public func expect<ExpectedWorkerType: Worker>(
         worker: ExpectedWorkerType,
         producingOutput output: ExpectedWorkerType.Output? = nil,
+        key: String = "",
+        file: StaticString = #file, line: UInt = #line
+    ) -> RenderTester<WorkflowType> {
+        expectWorker(worker, mockingOutput: output, key: key, file: file, line: line)
+    }
+
+    /// Mock the given worker's output, and assert that the workflow's worker `isEquivalent(to:)` the `expectedWorker`.
+    ///
+    /// - Parameters:
+    ///   - expectedWorker: The worker that we expect was created by the workflow. Will be compared using
+    ///                     `isEquivalent(to:)` to assert that the workflow's worker matches.
+    ///   - mockingOutput: The output to be used instead of actually running the worker.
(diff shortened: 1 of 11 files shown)


MigrationGuide_v1.0.md
```diff
@@ -120,9 +120,9 @@ NameLoadingWorkflow()
 // After
 NameLoadingWorkflow()
     .renderTester(initialState: .init(state: .loading, token: "user-token"))
-    .expect(
-        worker: LoadingWorker(token: "user-token"),
-        producingOutput: .success("Ben Cochran")
+    .mockWorker(
+        expectedWorker: LoadingWorker(token: "user-token"),
+        mockingOutput: .success("Ben Cochran")
```
jamieQ, line 125: i assume this was from a blanked find/replace, but probably no need to update this doc

WorkflowCombine/Testing/WorkerTesting.swift
```diff
@@ -21,29 +21,49 @@ import XCTest
...
+
+    /// Mock the given worker's output, and assert that the workflow's worker `isEquivalent(to:)` the `expectedWorker`.
+    ///
+    /// - Parameters:
+    ///   - expectedWorker: The worker that we expect was created by the workflow. Will be compared using
+    ///                     `isEquivalent(to:)` to assert that the workflow's worker matches.
+    ///   - mockingOutput: The output to be used instead of actually running the worker.
+    ///                    If the workflow never tries to run the worker, then this won't be used.
+    ///   - key: Key to expect this `Workflow` to be rendered with.
+    public func mockWorker<ExpectedWorkerType: Worker>(
```
jamieQ, line 50: per the feedback & discussions on naming we've had, shall we update this to `expect(worker: mockingOutput: ...)` or `expectWorker(_: mockingOutput: ...)`

WorkflowCombine/Testing/WorkerTesting.swift
```diff
@@ -21,29 +21,49 @@ import XCTest
...
     ///
     /// - Parameters:
-    ///   - worker: The worker to be expected
-    ///   - producingOutput: An output that will be returned when this worker is requested, if any.
+    ///   - worker: The worker that we expect was created by the workflow. Will be compared using
+    ///             `isEquivalent(to:)` to assert that the workflow's worker matches.
+    ///   - producingOutput: The output to be used instead of actually running the worker.
+    ///                      If the workflow never tries to run the worker, then this won't be used.
     ///   - key: Key to expect this `Workflow` to be rendered with.
+    @available(*, deprecated, renamed: "mockWorker", message: "This method's name & parameter names were unclear. It has been renamed to mockWorker.")
```
jamieQ, line 32: nit: the message is probably fine to just say what it has been renamed to without any further motivation

WorkflowCombine/Testing/WorkerTesting.swift
amayers: I went with `expectWorker`, as that caused less issues with the deprecation warning.

WorkflowCombine/Testing/WorkerTesting.swift
```diff
@@ -21,29 +21,49 @@ import XCTest
...
     ///
     /// - Parameters:
-    ///   - worker: The worker to be expected
-    ///   - producingOutput: An output that will be returned when this worker is requested, if any.
+    ///   - worker: The worker that we expect was created by the workflow. Will be compared using
+    ///             `isEquivalent(to:)` to assert that the workflow's worker matches.
+    ///   - producingOutput: The output to be used instead of actually running the worker.
+    ///                      If the workflow never tries to run the worker, then this won't be used.
     ///   - key: Key to expect this `Workflow` to be rendered with.
+    @available(*, deprecated, renamed: "expectWorker", message: "Renamed")
```
jamieQ, line 32: in these deprecation annotations, i think we likely want the `renamed:` parameter to fully specify the new method signature/param names so that fix-its can automatically be applied to update the relevant code. e.g. something like:

`renamed: "expectedWorker(_:mockingOutput:key:...)"`

though i didn't actually do a full integration test of this, it does seem to work locally to update existing callsites correctly when applying a fix-it.

PR #324, opened 2025-03-11
[gardening]: remove unneeded code from ChildKey

jamieQ:
small refactor of `ChildKey` to remove custom `Hashable` requirement implementations in favor of synthesized ones

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -404,22 +404,15 @@ extension WorkflowNode.SubtreeManager {
 
 extension WorkflowNode.SubtreeManager {
     struct ChildKey: Hashable {
-        var childType: Any.Type
+        var childTypeID: ObjectIdentifier
         var key: String
 
-        init(childType: (some Any).Type, key: String) {
-            self.childType = childType
+        init(
+            childType: (some Workflow).Type,
+            key: String
+        ) {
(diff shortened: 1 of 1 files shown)


Workflow/Sources/SubtreeManager.swift
```diff
@@ -404,23 +404,16 @@ extension WorkflowNode.SubtreeManager {
...
-
-        func hash(into hasher: inout Hasher) {
-            hasher.combine(ObjectIdentifier(childType))
-            hasher.combine(key)
-        }
-
-        static func == (lhs: ChildKey, rhs: ChildKey) -> Bool {
-            lhs.childType == rhs.childType
-                && lhs.key == rhs.key
-        }
```
jamieQ, line 423: we get these for free if all our stored properties are Hashable

mjohnson12 (approved):

thingdeux (approved):
Makes sense 🤙🏽 - LGTM

Workflow/Sources/SubtreeManager.swift
watt: you may find [this wrapper](https://github.com/square/Blueprint/blob/main/BlueprintUI/Sources/Internal/Metatype.swift) useful to do the same thing with better debugging/logging

watt (approved):

PR #325, opened 2025-03-12
[refactor]: introduce a 'HostContext' for exposing host info through tree

jamieQ:
- introduce a `HostContext` type to hold 'host-level' information to be distributed to various entities throughout the workflow tree
- currently it holds the two optional top-level injected properties: 'observers' and the 'debugger'
- in the future i expect this to be useful for propagating additional top-level configuration information downward (e.g. new runtime behavior options), as well as offering a convenient means for communicating back to the 'root' in some circumstances (e.g. recording nodes that changed when applying actions)

note: no new tests added as no new functionality introduced yet

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -36,14 +36,19 @@ extension WorkflowNode {
 
         private let session: WorkflowSession
 
-        private let observer: WorkflowObserver?
+        /// Reference to the context object for the entity hosting the corresponding node.
+        private let hostContext: HostContext
+
+        private var observer: WorkflowObserver? {
+            hostContext.observer
+        }
 
         init(
             session: WorkflowSession,
-            observer: WorkflowObserver? = nil
+            hostContext: HostContext
         ) {
             self.session = session
-            self.observer = observer
+            self.hostContext = hostContext
         }
 
         /// Performs an update pass using the given closure.
@@ -60,8 +65,8 @@ extension WorkflowNode {
                 previousSinks: previousSinks,
                 originalChildWorkflows: childWorkflows,
                 originalSideEffectLifetimes: sideEffectLifetimes,
-                session: session,
-                observer: observer
+                hostContext: hostContext,
+                session: session
             )
 
             let wrapped = RenderContext.make(implementation: context)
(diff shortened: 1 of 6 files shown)


Workflow/Sources/SubtreeManager.swift
```diff
@@ -194,7 +202,7 @@ extension WorkflowNode.SubtreeManager {
             let eventPipe = EventPipe()
             eventPipes.append(eventPipe)
 
-            /// See if we can
+            /// See if we can reuse an existing child node for the given key.
```
jamieQ, line 205: this truncated sentence has bothered me for a long time – i think this is probably what it was going to

mjohnson12 (approved):

kyleve (approved):

Workflow/Sources/WorkflowNode.swift
```diff
@@ -80,7 +86,7 @@ final class WorkflowNode<WorkflowType: Workflow> {
 
         switch subtreeOutput {
         case .update(let action, let source):
-            /// 'Opens' the existential `any WorkflowAction<WorkflowType>` value
+            /// 'Opens' the existential `any WorkfloswAction<WorkflowType>` value
```
watt, line 89: typo
```suggestion
            /// 'Opens' the existential `any WorkflowAction<WorkflowType>` value
```

Workflow/Sources/WorkflowHost.swift
```diff
@@ -115,3 +123,20 @@ public final class WorkflowHost<WorkflowType: Workflow> {
         outputEvent
     }
 }
+
+// MARK: - HostContext
+
+/// A context object to expose certain root-level information to each node
+/// in the Workflow tree.
+final class HostContext {
```
watt, line 131: class?

watt (approved):

Workflow/Sources/WorkflowHost.swift
jamieQ: sorry, are you asking why is it a class? assuming 'yes', then it's because i think we will eventually want to allow different parts of the tree to be able to mutate some of its properties and have those changes be visible elsewhere. e.g. you're handling an action and want to remember that 'some node in the tree was invalidated' by the time the root/host has to do the next render.

PR #327, opened 2025-03-13
Bring Async Worker syntax closer to standard Swift syntax

loganmoseley:
Hello! Some people have much more Workflow experience than I do, so feedback is welcome and appreciated.

# Why

To bring Async Worker syntax closer to standard Swift syntax, like a do-catch block with an await or try-await function call.

Example source: https://www.avanderlee.com/swift/async-await/
```swift
do {
    let images = try await fetchImages()
    print("Fetched \(images.count) images.")
} catch {
    print("Fetching images failed with error \(error)")
}
```
What if we could write something like this in Workflow?
```swift
context.run {
    let images = try await fetchImages(id)
    Action.updateResult(.success(images))
} catch {
    Action.updateResult(.failure(.notAvailable(error))
}
```

# What

1. Adds an `AsyncThrowingWorker` for convenient `async throws` function calling. It’s just like `AsyncOperationWorker`, but throwing.
2. Added a small group of convenience methods that basically flip `Worker.rendered(in:key:)` to `RenderContext.run(key:block:)`

Now we can write

### await

```swift
context.run(key: personID) {
    Action.updateList(await fetchFamily(personID))
}
```

### try await

Here are three things you could do.

1. If you want to keep your Action with Result:

```swift
context.run(key: id) {
    do {
        let images = try await fetchImages()
        Action.updateResult(.success(images))
    } catch {
        Action.updateResult(.failure(.notAvailable(error))
    }
}
```

2. If you want a little less indentation, but still a familiar “catch”, and you know you want to run it exactly once:

```swift
context.run {
    let images = try await fetchImages(id)
    Action.updateResult(.success(images))
} catch {
    Action.updateResult(.failure(.notAvailable(error))
}
```

3. Or if you just don’t really care about the error type, and you know you want to run it exactly once, get terse:

```swift
context.run {
    Action.updateResult(try await fetchImages(id))
}
```

# Checklist

- [x] Unit Tests
- ~UI Tests~
- ~Snapshot Tests (iOS only)~
    - I’m not sure about this. Is there more than the doc comments I’ve written?


WorkflowConcurrency/Sources/AsyncThrowingWorker.swift
```diff
@@ -0,0 +1,65 @@
...
+///         .running(in: context, key: "UniqueKey")
+///
+///     return MyScreen()
+/// }
+/// ```
+///
+/// Example of using a closure.
+/// ```
+/// func render(state: State, context: RenderContext<Self>) -> MyScreen {
+///     AsyncOperationWorker {
```
blakemcanally, line 36: docs reference `AsyncOperationWorker` and not `AsyncThrowingWorker`.

WorkflowConcurrency/Sources/AsyncOperationWorker.swift
```diff
@@ -60,3 +59,85 @@ public struct AsyncOperationWorker<OutputType>: Worker {
...
+                switch result {
+                case .success(let output):
+                    return output
+                case .failure(let error):
+                    return `catch`(error)
+                }
+            }
+    }
+
+    public func run<Action>(
```
blakemcanally, line 127: You'll want to document the overloads as well so that Xcode will pick up the docs in each case. Also, you should consider adding some code examples similar to what is on `AsyncThrowingWorker`

WorkflowConcurrency/Sources/AsyncOperationWorker.swift
```diff
@@ -60,3 +59,85 @@ public struct AsyncOperationWorker<OutputType>: Worker {
...
+    ///   - block: Any async closure or function to run. The return `Action`
+    ///   type must match the context's `WorkflowType`.
+    public func run<Action>(
+        keyFile: StaticString = #file,
+        keyFunction: StaticString = #function,
+        keyLine: Int = #line,
+        block: @escaping () async -> Action
+    ) where Action: WorkflowAction, Action.WorkflowType == WorkflowType
+    {
+        AsyncOperationWorker(block)
```
blakemcanally, line 84: nit: It might be worth implementing this method in terms of `run(key:block:)`

mjohnson12: +1

WorkflowConcurrency/Sources/AsyncOperationWorker.swift
```diff
@@ -60,3 +59,85 @@ public struct AsyncOperationWorker<OutputType>: Worker {
...
+
+    public func run<Action>(
+        keyFile: StaticString = #file,
+        keyFunction: StaticString = #function,
+        keyLine: Int = #line,
+        block: @escaping () async throws -> Action,
+        catch: @escaping (Error) -> Action
+    ) where Action: WorkflowAction, Action.WorkflowType == WorkflowType
+    {
+        AsyncThrowingWorker(block)
```
mjohnson12, line 116: Same as Blake's comment above.  This can just call `run(key:block:)`

WorkflowConcurrency/Sources/AsyncOperationWorker.swift
```diff
@@ -60,3 +59,85 @@ public struct AsyncOperationWorker<OutputType>: Worker {
...
+        keyFile: StaticString = #file,
+        keyFunction: StaticString = #function,
+        keyLine: Int = #line,
+        block: @escaping () async -> Output,
+        map: @escaping (Output) -> Action
+    ) where Action: WorkflowAction, Action.WorkflowType == WorkflowType
+    {
+        AsyncOperationWorker(block)
+            .running(in: self, key: "\(keyFile).\(keyFunction):\(keyLine)", outputMap: map)
+    }
```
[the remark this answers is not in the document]
mjohnson12: I'm for removing it.

WorkflowConcurrency/Sources/AsyncThrowingWorker.swift
```diff
@@ -0,0 +1,65 @@
...
+    public func run() async -> Result<Success, Error> {
+        do {
+            return .success(try await operation())
+        } catch {
+            return .failure(error)
+        }
+    }
+
+    public func isEquivalent(to otherWorker: AsyncThrowingWorker<Success>) -> Bool {
+        true
```
kyleve, line 63: Dumb workflow question, but is `true` always right here? Eg what if the closure passed changes?

loganmoseley: I don’t know. This is what `AsyncOperationWorker` does.

loganmoseley:
Closing this for now, until I get the kinks hammered out.

PR #329, opened 2025-03-14
[gardening]: fix some deprecation warnings

jamieQ:
updates some testing API to use more defaulted parameters and pass them through to the `expectNoDifference` method from `CustomDump`, as the existing `XCTAssertNoDifference` has been deprecated.

### DIFF
--- a/WorkflowTesting/Sources/RenderTesterResult.swift
+++ b/WorkflowTesting/Sources/RenderTesterResult.swift
@@ -156,10 +156,10 @@ extension RenderTesterResult where WorkflowType.State: Equatable {
     @discardableResult
     public func assertStateModifications(
         file: StaticString = #file,
-        line: UInt = #line,
-        _ modifications: (inout WorkflowType.State) throws -> Void,
         fileID: StaticString = #fileID,
-        column: UInt = #column
+        line: UInt = #line,
+        column: UInt = #column,
+        _ modifications: (inout WorkflowType.State) throws -> Void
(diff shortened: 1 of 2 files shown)


WorkflowTesting/Sources/RenderTesterResult.swift
```diff
@@ -144,16 +157,20 @@ extension RenderTesterResult where WorkflowType.State: Equatable {
     public func assertStateModifications(
         file: StaticString = #file,
         line: UInt = #line,
-        _ modifications: (inout WorkflowType.State) throws -> Void
+        _ modifications: (inout WorkflowType.State) throws -> Void,
+        fileID: StaticString = #fileID,
+        column: UInt = #column
```
jamieQ, line 162: this isn't ideal... should the new params go before or after the 'real' parameter? generally i think defaulted stuff is supposed to come at the end, but... this was seemingly never implemented that way. would consistency be better in this case?

kyleve: Default params should usually go at the end, _but_ IIRC these should go before the trailing closures to preserve that syntax?

jamieQ: since they are new i think it works either way but... being consistent is probably the right thing to do. we could perhaps unify everything and deprecate the old stuff and it will 'just work' if these aren't being passed in explicitly anyway.

nabs-m (approved):

PR #331, opened 2025-03-26
[perf]: don't instantiate debugger info if not needed

jamieQ:
refactors action propagation to no longer instantiate any `WorkflowDebugInfo` types unless there is a `WorkflowDebugger` set on the host. this should avoid some unnecessary string allocations and reflection metadata lookups.

## Checklist

- [x] Unit Tests
- [x] UI Tests

### DIFF
--- a/Workflow/Sources/Debugging.swift
+++ b/Workflow/Sources/Debugging.swift
@@ -161,7 +161,7 @@ extension WorkflowUpdateDebugInfo? {
 
 extension WorkflowUpdateDebugInfo {
     fileprivate static let unexpectedlyMissing = {
-        assertionFailure("Creation of actual WorkflowUpdateDebugInfo failed unexectedly")
+        assertionFailure("Creation of actual WorkflowUpdateDebugInfo failed unexpectedly")
         return WorkflowUpdateDebugInfo(
             workflowType: "BUG IN WORKFLOW",
             kind: .didUpdate(source: .external)
--- a/Workflow/Sources/Debugging.swift
+++ b/Workflow/Sources/Debugging.swift
@@ -148,3 +148,23 @@ extension WorkflowHierarchyDebugSnapshot {
         }
     }
 }
+
+// MARK: - Compatibility
+
+/// These extensions are utilities to support conditionally emitting debug info only when a
+/// `debugger` is set.
+extension WorkflowUpdateDebugInfo? {
+    var unwrappedOrErrorDefault: WorkflowUpdateDebugInfo {
+        self ?? .unexpectedlyMissing
+    }
+}
+
+extension WorkflowUpdateDebugInfo {
+    fileprivate static let unexpectedlyMissing = {
+        assertionFailure("Creation of actual WorkflowUpdateDebugInfo failed unexectedly")
+        return WorkflowUpdateDebugInfo(
+            workflowType: "BUG IN WORKFLOW",
+            kind: .didUpdate(source: .external)
+        )
+    }()
+}
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -131,10 +131,29 @@ extension WorkflowNode {
     }
 }
 
+/// Carries information about the provenance of an event handled by the runtime.
+enum EventSource: Equatable {
+    /// An event received from an external source.
+    case external
+
+    /// An event that from a descendent in the subtree.
+    /// Will contain associated debug info iff the `debugger` property of the `WorkflowHost`
+    /// is set.
+    case subtree(WorkflowUpdateDebugInfo?)
+
+    /// Compatibility method to convert to the public representation of this data
+    func toDebugInfoSource() -> WorkflowUpdateDebugInfo.Source {
+        switch self {
+        case .external: .external
+        case .subtree(let maybeInfo): .subtree(maybeInfo.unwrappedOrErrorDefault)
+        }
+    }
+}
+
 extension WorkflowNode.SubtreeManager {
     enum Output {
-        case update(any WorkflowAction<WorkflowType>, source: WorkflowUpdateDebugInfo.Source)
-        case childDidUpdate(WorkflowUpdateDebugInfo)
+        case update(any WorkflowAction<WorkflowType>, source: EventSource)
+        case childDidUpdate(WorkflowUpdateDebugInfo?)
     }
 }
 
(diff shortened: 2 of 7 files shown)


Workflow/Sources/WorkflowHost.swift
```diff
@@ -95,10 +95,12 @@ public final class WorkflowHost<WorkflowType: Workflow> {
         // Treat the update as an "output" from the workflow originating from an external event to force a render pass.
         let output = WorkflowNode<WorkflowType>.Output(
             outputEvent: nil,
-            debugInfo: WorkflowUpdateDebugInfo(
-                workflowType: "\(WorkflowType.self)",
-                kind: .didUpdate(source: .external)
-            )
+            debugInfo: context.ifDebuggerEnabled {
```
kyleve, line 98: clever!

kyleve (approved):

Workflow/Sources/SubtreeManager.swift
```diff
@@ -131,10 +131,29 @@ extension WorkflowNode {
...
+    /// Will contain associated debug info iff the `debugger` property of the `WorkflowHost`
+    /// is set.
+    case subtree(WorkflowUpdateDebugInfo?)
+
+    /// Compatibility method to convert to the public representation of this data
+    func toDebugInfoSource() -> WorkflowUpdateDebugInfo.Source {
+        switch self {
+        case .external: .external
+        case .subtree(let maybeInfo): .subtree(maybeInfo.unwrappedOrErrorDefault)
+        }
```
jamieQ, line 149: the other two cases of debug info source (`worker` and `sideeffect`) seemed to never have been provided here AFAICT, so we only bother representing the other two cases in the new type.

Workflow/Sources/SubtreeManager.swift
```diff
@@ -131,10 +131,29 @@ extension WorkflowNode {
...
+    }
+}
+
 extension WorkflowNode.SubtreeManager {
     enum Output {
-        case update(any WorkflowAction<WorkflowType>, source: WorkflowUpdateDebugInfo.Source)
-        case childDidUpdate(WorkflowUpdateDebugInfo)
+        case update(any WorkflowAction<WorkflowType>, source: EventSource)
```
jamieQ, line 155: changed this from using the public type to a new internal one

Workflow/Sources/Debugging.swift
```diff
@@ -148,3 +148,23 @@ extension WorkflowHierarchyDebugSnapshot {
...
+/// These extensions are utilities to support conditionally emitting debug info only when a
+/// `debugger` is set.
+extension WorkflowUpdateDebugInfo? {
+    var unwrappedOrErrorDefault: WorkflowUpdateDebugInfo {
+        self ?? .unexpectedlyMissing
+    }
+}
+
+extension WorkflowUpdateDebugInfo {
+    fileprivate static let unexpectedlyMissing = {
```
jamieQ, line 163: in a few places we want to propagate debug info with the 'right' optionality. either it should be `nil` if there's no debugger, or non-`nil` if there is. at the root level we need to emit a value to the debugger if present that is non-optional, hence this 'sanity check' value that will hopefully ensure we don't mess things up.

jamieQ (commented):
made these changes in a non-breaking way. IMO the current manifestation of the `WorkflowDebugger` API has not seemed to provide a ton of value for us internally. it might be worth considering removing it in the next major version bump, which would allow some of this code to be cleaned up.

Workflow/Sources/Debugging.swift
blakemcanally: Is there an integration test we could write to validate this behavior, or some other mechanism to solidify this contract?

Workflow/Sources/Debugging.swift
```diff
@@ -148,3 +148,23 @@ extension WorkflowHierarchyDebugSnapshot {
...
+        assertionFailure("Creation of actual WorkflowUpdateDebugInfo failed unexectedly")
```
square-tomb, line 164:
```suggestion
        assertionFailure("Creation of actual WorkflowUpdateDebugInfo failed unexpectedly")
```

Workflow/Sources/SubtreeManager.swift
```diff
@@ -131,10 +131,29 @@ extension WorkflowNode {
     }
 }
 
+/// Carries information about the provenance of an event handled by the runtime.
+enum EventSource: Equatable {
+    /// An event received from an external source.
+    case external
+
+    /// An event that from a descendent in the subtree.
```
square-tomb, line 139:
```suggestion
    /// An event that comes from a descendent in the subtree.
```

square-tomb (approved):

Workflow/Sources/Debugging.swift
jamieQ: i added a few more unit tests to cover the 'set a debugger' and 'didn't set a debugger' cases

blakemcanally (approved):

Workflow/Tests/HostContextTests.swift
```diff
@@ -0,0 +1,23 @@
...
+    func test_conditional_debug_info_no_debugger() {
+        let subject = HostContext.testing(debugger: nil)
+        subject.ifDebuggerEnabled {
+            XCTFail("should not be called")
+        }
+    }
+
+    func test_conditional_debug_info_with_debugger() {
+        let subject = HostContext.testing(debugger: TestDebugger())
+        let expectaiton = expectation(description: "debugger block invoked")
```
watt, line 15:
```suggestion
        let expectation = expectation(description: "debugger block invoked")
```
Did you know you can turn on spellcheck in Xcode?

watt (approved):
> this should avoid some unnecessary string allocations and reflection metadata lookups

Did you verify that there's a performance improvement, or is this theoretical?

Workflow/Tests/HostContextTests.swift
jamieQ: ha... yes i do, and (perhaps surprisingly) do in fact have that setting enabled. guess i was not paying very close attention...

jamieQ:
> Did you verify that there's a performance improvement, or is this theoretical?

it's largely theoretical, and i'd expect any 'real world' gains to be small and amortized, but the code we're currently always running does call runtime functions to produce strings from types, which is not the cheapest thing ever. i did a rough sanity check locally when applying an action through a nested chain of workflows 1000 nodes long, and it appeared consistently faster to skip the debug info creation if not needed.

PR #333, opened 2025-04-04
feat: nested observable state stores

watt:
This PR adds new scoping methods to `Store` that allow you to create child stores from nested observable state.

These same scoping methods already exist for child _models_, typically created by rendering child workflows and composing their renderings, but not for nested state within a single workflow.

I've added a new `ObservableCounter` sample to demonstrate. The existing `ObservableScreen` sample has been renamed to `ObservableComposition`.

## Example

Consider a state like this:

```swift
@ObservableState
struct State {
  var children: [ChildState]

  @ObservableState
  struct ChildState: Identifiable {
    let id = UUID()
    var foo: Int
  }
}
```

If one was trying to iterate over `children` and form bindings to the `foo` property, you might attempt this construction, which creates a `Binding<[ChildState]>` and then uses `ForEach` to map over individual `Binding<ChildState>`s.

```swift
struct ExampleView: View {
  @Perception.Bindable
  var store: Store<StateAccessor<State>>

  var body: some View {
    ForEach($store.children) { child in
      FooView(foo: child.foo)
    }
  }
}
```

Unfortunately this will compile and run, but writes to `child.foo` will not cause `FooView` to be invalidated. The exact mechanics behind this are opaque, but the bindings captured by the `ForEach` don't register observation at the right scope.

Using the new scoping methods, you can iterate over child stores, and create a binding from each store instead.

```swift
struct ExampleView: View {
  var store: Store<StateAccessor<State>>

  var body: some View {
    ForEach(store.scope(collection: \.children)) { child in
      // you can also inline this incantation to the capture list if you prefer
      @Perception.Bindable var child = child
      FooView(foo: $child.foo)
    }
  }
}
```

## Checklist

- [x] Unit Tests
- [x] UI Tests


Samples/ObservableCounter/Sources/CounterListView.swift
```diff
@@ -0,0 +1,23 @@
+import SwiftUI
+import WorkflowSwiftUI
+
+struct CounterListView: View {
+    @Perception.Bindable
+    var store: Store<CounterListModel>
+
+    var body: some View {
+        let _ = print("CounterListView.body")
```
watt, line 9: The printing is intentionally left in these samples to show the effect of state changes on body evaluations.

WorkflowSwiftUI/Sources/Derived/ObservableState.swift
```diff
@@ -113,6 +113,20 @@ public func _$isIdentityEqual<T: ObservableState>(
     lhs._$id == rhs._$id
 }
 
+@inlinable
+public func _$isIdentityEqual<T: ObservableState>(
```
watt, line 117: This addition is not strictly needed for scoping, but makes the optional case much more efficient when going from nil to nil.

WorkflowSwiftUI/Tests/Derived/ObservableStateTests.swift
```diff
@@ -140,16 +140,14 @@ final class ObservableStateTests: XCTestCase {
         // nil -> nil
         do {
             var state = ParentState(optional: nil)
-            let optionalDidChange = expectation(description: "optional.didChange")
 
             withPerceptionTracking {
                 _ = state.optional
             } onChange: {
-                optionalDidChange.fulfill()
+                XCTFail("Optional should not change")
```
watt, line 147: result of the new `_$isIdentityEqual` overload

Package.resolved
watt, line 1: The Tuist setup means this file at best redundant and at worst completely out of sync. (I can pull this into a separate PR if it's controversial.)

nsillik (commented):
Nice addition, surprisingly easy. 

I was wondering if it would be possible to create a purple warning when folks try to do this the wrong way. Since we've already had people try to reach for `ForeEach(store.children)` substates, we'll definitely want to encourage them to use scoped stores (and aren't likely to catch 100% of people who don't!). A runtime warning (like you get when you forget `WithPerceptionTracking`) seems like the easiest way to go (or an `assert` if you're feeling especially punitive, I suppose)

watt:
> I was wondering if it would be possible to create a purple warning when folks try to do this the wrong way. Since we've already had people try to reach for `ForeEach(store.children)` substates, we'll definitely want to encourage them to use scoped stores (and aren't likely to catch 100% of people who don't!). A runtime warning (like you get when you forget `WithPerceptionTracking`) seems like the easiest way to go (or an `assert` if you're feeling especially punitive, I suppose)

It might be, but it's a little tricky. I'm not sure we can do it reliably, and there are cases where it's still OK to access nested state without a store. I think I want to wait and see how usage develops first.

johnnewman-square (approved):
It’s great to have a working solution to this one. The sample app provides good context, too.

Samples/ObservableCounter/Sources/CounterListView.swift
```diff
@@ -0,0 +1,23 @@
...
+        WithPerceptionTracking {
+            VStack {
+                ForEach(store.scope(collection: \.counters)) { counter in
+                    @Perception.Bindable var counter = counter
+
+                    WithPerceptionTracking {
```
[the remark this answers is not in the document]
johnnewman-square: Removing that `WithPerceptionTracking ` triggers those Perception runtime warnings. Likely because of the binding to the child store used inside the `ForEach` escaping closure.

Samples/ObservableCounter/Sources/CounterListView.swift
```diff
@@ -0,0 +1,23 @@
...
+struct CounterListView: View {
+    @Perception.Bindable
+    var store: Store<CounterListModel>
+
+    var body: some View {
+        let _ = print("CounterListView.body")
+        WithPerceptionTracking {
+            VStack {
+                ForEach(store.scope(collection: \.counters)) { counter in
+                    @Perception.Bindable var counter = counter
```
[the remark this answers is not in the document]
watt: > I'm a little surprised this works – maybe just stable because counters isn't changing?

You mean the shadowing, or the fact that we can wrap it at all? The pattern for a local `@Bindable` actually [comes straight from Apple](https://developer.apple.com/documentation/swiftui/bindable).

> Is it possible to have the scope collection method vend a bindable variable, or does that not work?

That won't work, or more accurately, it's uglier, because you'd lose the property wrapper magic and have to call into `counter.wrappedValue` or `counter.projectedValue` manually.

Samples/ObservableCounter/Sources/CounterListView.swift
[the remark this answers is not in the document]
watt: Yep, `ForEach` is a case that requires nested `WithPerceptionTracking`.

Samples/ObservableCounter/Sources/CounterListView.swift
```diff
@@ -0,0 +1,23 @@
+import SwiftUI
```
jamieQ, line 1: commenting here primarily for threading purposes, but wondering if this issue is specific to our setup, or can something analogous be replicated with 'vanilla' SwiftuI?

WorkflowSwiftUI/Tests/NestedStoreTests.swift
```diff
@@ -0,0 +1,431 @@
...
+    func test_nested_optional() async {
+        // some to nil
+        do {
+            var state = State(optional: .init())
+            let model = StateAccessor(state: state) { update in
+                update(&state)
+            }
+            let (store, _) = Store.make(model: model)
+
+            state.optional = .init()
```
square-tomb, line 35: nit: Is this necessary given `state` was initialized with `optional: .init()`?

square-tomb (approved):

Samples/ObservableCounter/Sources/CounterListView.swift
watt: Specific to our setup. There may be some assumptions built around the state being a reference type that we're running into.

WorkflowSwiftUI/Tests/NestedStoreTests.swift
watt: nope, removed

WorkflowSwiftUI/Sources/Derived/ObservableState.swift
robmaceachern: Is it worth moving this to a file outside of the `Derived` folder? It'd probably make diffs between this and the TCA ObservableState easier.

Documentation/WorkflowSwiftUI Adoption Guide.md
```diff
@@ -341,6 +343,81 @@ struct CoupleView: View {
...
+struct State {
+
+  @ObservableState
+  struct CounterState: Identifiable {
+    let id = UUID()
+    var count = 0
+  }
+
+  var counter = CounterState()
+  var optionalCounter? = CounterState()
```
robmaceachern, line 363:
```suggestion
  var optionalCounter: CounterState? = CounterState()
```

robmaceachern (approved):

WorkflowSwiftUI/Sources/Derived/ObservableState.swift
watt: 🤷  previously [I was asked](https://github.com/square/workflow-swift/pull/283#discussion_r1681291606) to put it _in_ this folder for the same reason.

robmaceachern: I just mean this function specifically (if it's not from the upstream repo)

WorkflowSwiftUI/Sources/Store.swift
```diff
@@ -262,13 +304,32 @@ extension Store {
...
     }
 
-    // Optionals
+    /// Derives a store for nested observable state on the model.
+    ///
+    /// - Parameter keyPath: the keypath to the substate
+    /// - Returns: a store
+    public func scope<Substate: ObservableState>(
+        keyPath: WritableKeyPath<Model.State, Substate>
+    ) -> Store<StateAccessor<Substate>> {
```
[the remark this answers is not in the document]
watt: Not that I'm aware of, unfortunately. I'm pleasantly surprised that the error message is decent.

PR #334, opened 2025-04-04
[perf]: a couple more micro optimizations

jamieQ:
small changes to:

1. avoid an unnecessary buffer copy during render() when updating event pipes
2. allow our 'session' id generation to wrap so it doesn't need to perform bounds checks (it functionally cannot overflow so they are unneeded)

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -82,15 +82,19 @@ extension WorkflowNode {
             childWorkflows = context.usedChildWorkflows
             sideEffectLifetimes = context.usedSideEffectLifetimes
 
-            /// Captured the reusable sinks from this render pass.
+            /// Capture the reusable sinks from this render pass.
             previousSinks = context.sinkStore.usedSinks
 
             /// Capture all the pipes to be enabled after render completes.
             eventPipes = context.eventPipes
-            eventPipes.append(contentsOf: context.sinkStore.eventPipes)
+            for (_, sink) in context.sinkStore.usedSinks {
+                eventPipes.append(sink.eventPipe)
+            }
 
             /// Set all event pipes to `pending`.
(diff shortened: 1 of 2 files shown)


Workflow/Sources/SubtreeManager.swift
```diff
@@ -82,15 +82,19 @@ extension WorkflowNode {
...
             /// Capture all the pipes to be enabled after render completes.
             eventPipes = context.eventPipes
-            eventPipes.append(contentsOf: context.sinkStore.eventPipes)
+            for (_, sink) in context.sinkStore.usedSinks {
+                eventPipes.append(sink.eventPipe)
+            }
 
             /// Set all event pipes to `pending`.
-            eventPipes.forEach { $0.setPending() }
+            for eventPipe in eventPipes {
```
jamieQ, line 95: not really a functional change here, but we use for-in everywhere else in this method so we might as well be consistent

Workflow/Sources/SubtreeManager.swift
```diff
@@ -82,15 +82,19 @@ extension WorkflowNode {
...
             sideEffectLifetimes = context.usedSideEffectLifetimes
 
-            /// Captured the reusable sinks from this render pass.
+            /// Capture the reusable sinks from this render pass.
             previousSinks = context.sinkStore.usedSinks
 
             /// Capture all the pipes to be enabled after render completes.
             eventPipes = context.eventPipes
-            eventPipes.append(contentsOf: context.sinkStore.eventPipes)
+            for (_, sink) in context.sinkStore.usedSinks {
```
jamieQ, line 90: previously this was going through a call to `map` which created an intermediate buffer of the mapped elements. we don't need them to be stored separately so we can just do the iteration inline to avoid that

mjohnson12 (approved):

Issue #339, opened 2025-05-02
Add a convenience for mapping outputs from `Never` to any type that does not produce compiler warnings


jamieQ:
today, attempting to do this both requires an instance of the new type and produces a compiler warning. we should offer a special case for this that makes the type signatures 'just work' without requiring a value and doesn't emit a compiler warning. example:

```swift
struct WF: Workflow {
    typealias Output = Never
    typealias State = Void
    typealias Rendering = Int

    func render(state: State, context: RenderContext<WF>) -> Rendering {

        return 0
    }
}

func test(_ rc: RenderContext<WF>) {
    enum NewOutput {
        case zero
    }

    let z: AnyWorkflow<String, NewOutput> = WF()
        .mapRendering { "\($0)" }
        .mapOutput { _ in .zero } // ⚠️ Will never be executed
        .asAnyWorkflow()

}
```

PR #340, opened 2025-05-02
Adds a convenience for mapping Never outputs to other types

jszumski:
Adds a way to map a workflow's `Never` output type to a different type without generating compiler warnings.  Fixes https://github.com/square/workflow-swift/issues/339

I like this solution because the natural inclination is to simply delete the block containing unreachable code and that just works:

```diff
-            .mapOutput { switch $0 {} }
+            .mapOutput()
```

I'm happy to add a test using this but the true validation is to successfully build without warnings.

### DIFF
--- a/Workflow/Sources/AnyWorkflow.swift
+++ b/Workflow/Sources/AnyWorkflow.swift
@@ -195,3 +195,14 @@ extension AnyWorkflowConvertible {
         asAnyWorkflow().mapRendering(transform)
     }
 }
+
+// MARK: -
+
+extension AnyWorkflowConvertible where Output == Never {
+    /// A convenience for workflows that don't produce output but want to change the declared output type.
+    public func mapOutput<NewOutput>(to: NewOutput.Type) -> AnyWorkflow<Rendering, NewOutput> {
+        mapOutput(mapNever(_:))
+    }
+}
+
+private func mapNever<T>(_ input: Never) -> T {}
--- a/Workflow/Tests/AnyWorkflowTests.swift
+++ b/Workflow/Tests/AnyWorkflowTests.swift
@@ -82,6 +82,13 @@ public class AnyWorkflowTests: XCTestCase {
 
         XCTAssertNotNil(erased.base as? OnOutputWorkflow)
     }
+
+    func testMapNeverToOtherType() {
+        // test that this `Output = Bool` workflow compiles when mapped from an `Output = Never` workflow
+        let boolOutputWorkflow: AnyWorkflow<String, Bool> = NeverOutputWorkflow().mapOutput(to: Bool.self)
+
+        XCTAssertNotNil(boolOutputWorkflow)
+    }
 }
 
 /// Has no state or output, simply renders a reversed string
@@ -148,5 +155,15 @@ private struct OnOutputChildWorkflow: Workflow {
         DispatchQueue.main.async {
             sink.send(true)
         }
+    }
+}
+
+private struct NeverOutputWorkflow: Workflow {
+    typealias Output = Never
+    typealias Rendering = String
+    typealias State = Void
+
+    func render(state: State, context: RenderContext<NeverOutputWorkflow>) -> String {
+        "Never"
     }
 }


Workflow/Sources/AnyWorkflowConvertible.swift
```diff
@@ -147,3 +147,18 @@ struct OutputBlockingWorkflow<Child: AnyWorkflowConvertible>: Workflow {
             .rendered(in: context)
     }
 }
+
+// MARK: -
+
+extension AnyWorkflowConvertible where Output == Never {
```
jamieQ, line 153: i think this extension should be moved near [these existing ones](https://github.com/square/workflow-swift/blob/deb7e21e0c97fe4c9010d338efb24440e44e69a9/Workflow/Sources/AnyWorkflow.swift#L189).

Workflow/Sources/AnyWorkflowConvertible.swift
```diff
@@ -147,3 +147,18 @@ struct OutputBlockingWorkflow<Child: AnyWorkflowConvertible>: Workflow {
...
+    /// A convenience for workflows that don't produce output but want to change the declared output type.
+    ///
+    /// This equivalent to `.mapOutput { switch $0 {} }` but avoids the compiler warning for unreachable code in the provided transform.
```
jamieQ, line 156: i think it'd be okay to drop this from the public declaration docs. technically i think you'd need a larger expression to even tell how the type conversion is happening. seems like it would be a good thing to keep as an internal comment though.

Workflow/Sources/AnyWorkflowConvertible.swift
```diff
@@ -147,3 +147,18 @@ struct OutputBlockingWorkflow<Child: AnyWorkflowConvertible>: Workflow {
...
+    public func mapOutput<NewOutput>() -> AnyWorkflow<Rendering, NewOutput> {
+        asAnyWorkflow().mapOutput(silenceUnreachableWarning(_:))
```
jamieQ, line 158: nit: i think this could just call through to the `mapOutput` implementation without doing the `asAnyWorkflow()` call (since that will happen in the version that takes a transform parameter).

Workflow/Sources/AnyWorkflowConvertible.swift
```diff
@@ -147,3 +147,18 @@ struct OutputBlockingWorkflow<Child: AnyWorkflowConvertible>: Workflow {
...
     }
 }
+
+// MARK: -
+
+extension AnyWorkflowConvertible where Output == Never {
+    /// A convenience for workflows that don't produce output but want to change the declared output type.
+    ///
+    /// This equivalent to `.mapOutput { switch $0 {} }` but avoids the compiler warning for unreachable code in the provided transform.
+    public func mapOutput<NewOutput>() -> AnyWorkflow<Rendering, NewOutput> {
```
jamieQ, line 157: nit: mostly thinking 'out loud', but if for some reason type inference isn't 'working', i wonder if we might want to use a defaulted parameter here so it can be explicitly controlled? i think i convinced myself it's simpler as-is and we could always add that later, so no need to change anything.

Workflow/Sources/AnyWorkflowConvertible.swift
```diff
@@ -147,3 +147,18 @@ struct OutputBlockingWorkflow<Child: AnyWorkflowConvertible>: Workflow {
...
+        asAnyWorkflow().mapOutput(silenceUnreachableWarning(_:))
+    }
+
+    private func silenceUnreachableWarning<T>(
```
jamieQ, line 161: nit: let's move this out of the extension so it's not implicitly nested in a generic type, and rename it for the substance of what it's doing. also it seems the compiler allows you to implement this without a method body, so it can be reduced to something like:

```swift
private func mapNever<T>(_ input: Never) -> T {}
```

jamieQ (commented):
thanks! left some minor comments. also a test that this new method 'does the right thing' should be added, probably to `AnyWorkflowTests.swift`

Workflow/Sources/AnyWorkflowConvertible.swift
jszumski: Removed

Workflow/Sources/AnyWorkflowConvertible.swift
jszumski: Good call, much cleaner

jszumski:
Great feedback, I also added a simple test for it

square-tomb:
This reminds me of ReactiveSwift's [`promoteError`](https://github.com/ReactiveCocoa/ReactiveSwift/blob/master/Documentation/BasicOperators.md#promote). I notice theirs take the target type as an explicit parameter:
```swift
numbersSignal
    .promoteError(NSError.self)
```

I wonder if there are situations where that parameter is necessary for type inference to be succeed or be performant?

jamieQ (approved):
lgtm – thanks!

Issue #330, opened 2025-03-18
Add support for Swift Testing in test facilities


jamieQ:
today, test assertions are a combo of XCTest and utilities from `swift-custom-dump`, and they don't always work as expected when running tests via Swift Testing. make a pass through the code to ensure things are compatible regardless of which test framework is used.

PR #341, opened 2025-05-08
[chore]: improve Swift Testing support in test utilities

jamieQ:
migrates test utilities from `XCTest` assertions to analogs provided by the `CustomDump` and `IssueReporting` libraries. this should fix the issue where assertion failures would be invisible to client code written using Swift Testing rather than XCTest.

ideally we should probably add a lint check in the relevant places to enforce use of the appropriate API, but that will be a separate task.

resolves: https://github.com/square/workflow-swift/issues/330

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -65,6 +65,7 @@ let package = Package(
         .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.4.0"),
         .package(url: "https://github.com/pointfreeco/swift-perception", from: "1.5.0"),
         .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.2.1"),
+        .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.0.0"),
     ],
     targets: [
         // MARK: Workflow
--- a/WorkflowCombine/Testing/WorkerTesting.swift
+++ b/WorkflowCombine/Testing/WorkerTesting.swift
@@ -15,6 +15,7 @@
  */
 
 #if DEBUG
+import IssueReporting
 import Workflow
 import WorkflowTesting
 import XCTest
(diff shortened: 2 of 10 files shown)


WorkflowCombine/Testing/WorkerTesting.swift
```diff
@@ -15,6 +15,7 @@
  */
 
 #if DEBUG
+import IssueReporting
```
gyfelton, line 18: for my education: where is this lib coming from? Apple side?

jamieQ: this is a pointfree library – it's a dependency of the CustomDump lib that we already use: https://github.com/pointfreeco/swift-issue-reporting

jamieQ: though now that you bring it up... i wonder if it should be listed as an explicit dependency in the package?

gyfelton: yup and i see where the magic is coming from: https://github.com/pointfreeco/swift-issue-reporting/blob/main/Sources/IssueReporting/ReportIssue.swift#L71

Package.swift
```diff
@@ -65,6 +65,7 @@ let package = Package(
         .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.4.0"),
         .package(url: "https://github.com/pointfreeco/swift-perception", from: "1.5.0"),
         .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.2.1"),
+        .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.0.0"),
```
jamieQ, line 68: we currently already depend on this transitively through 'swift-custom-dump'. added as an explicit dependency because we're directly importing it in some spots now, and i assume that's the 'right thing to do'.

WorkflowCombine/Testing/WorkerTesting.swift
```diff
@@ -62,9 +63,9 @@ extension RenderTester {
                 guard !workflow.worker.isEquivalent(to: expectedWorker) else {
                     return
                 }
-                XCTFail(
+                reportIssue(
```
watt, line 66: There's no Apple-provided way to do this?

jamieQ: by 'this', what do you mean? record an issue in either testing framework?

mjohnson12 (approved):

PR #343, opened 2025-05-09
[chore]: update CI config

jamieQ:
updates the CI config with the following changes:

1. bump to Xcode 16.1
2. add a test matrix for the unit tests
    i. this is largely copied from the analogous logic in the Blueprint CI [config](https://github.com/square/Blueprint/blob/067015242d815f698550b3960353f545beea7183/.github/workflows/tests.yaml)
    ii. runs tests against the 3 most recent iOS SDKs (16.4, 17.5, 18.1)
3. split out snapshot tests to run only against iOS 18.1
    i. this was primarily due to not knowing how to configure them to function correctly with the test matrix – something that should presumably be improved, but is no worse than what we're currently doing.
4. disables a failing unit test after updating Xcode
    i. TODO: re-enable by adding support for app-hosted unit tests (tracked via [this issue](https://github.com/square/workflow-swift/issues/346))


WorkflowSwiftUI/Tests/PreferredContentSizeTests.swift
```diff
@@ -6,7 +6,12 @@ import WorkflowSwiftUI
 import XCTest
 
 final class PreferredContentSizeTests: XCTestCase {
-    func test_preferredContentSize() {
+    func test_preferredContentSize() throws {
+        guard #unavailable(iOS 18.0) else {
+            try XCTSkipIf(true, "Intrinsic content size does not update unless hosted in an actual app on iOS 18")
+            return
+        }
```
jamieQ, line 13: not sure what a better solution to this problem is (iOS 18 runtime seemingly doesn't invalidate/update VCs in the same way it used to)... use app hosted unit tests?

.github/workflows/swift.yaml
```diff
@@ -7,34 +7,75 @@ on:
...
+
+    - name: Run Tests
+      run: tuist test --path Samples ${{ matrix.scheme }} --device "${{ env.TUIST_TEST_DEVICE }}" --os "${{ matrix.sdk }}"
+
+  # FIXME: these should probably be run with a matrix too
+  snapshot-tests:
+    runs-on: macos-latest
+
+    strategy:
+      fail-fast: false # Don’t fail-fast so that we get all snapshot test changes
```
jamieQ, line 65: not sure this is actually needed, but copied it from the Blueprint config

WorkflowSwiftUI/Tests/PreferredContentSizeTests.swift
jamieQ: sounds good, but let's do as a separate PR then (unless it requires very little code change)

.github/workflows/swift.yaml
```diff
@@ -7,34 +7,75 @@ on:
...
+    steps:
+    - uses: actions/checkout@v4
+    - uses: jdx/mise-action@v2
+
+    - name: Switch to Xcode ${{ env.XCODE_VERSION }}
+      run: sudo xcode-select -s /Applications/Xcode_${{ env.XCODE_VERSION }}.app
+
+    - name: Install iOS ${{ matrix.sdk }}
+      if: ${{ matrix.installation_required }}
+      run: sudo xcodes runtimes install "iOS ${{ matrix.sdk }}"
```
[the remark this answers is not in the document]
jamieQ: what things have you seen break? personally i don't feel too strongly, but figure we might as well try hitting all the versions we support if it's not too hard

WorkflowSwiftUI/Tests/PreferredContentSizeTests.swift
jamieQ: this 'workaround' didn't actually seem to work around the problem for some reason, so i've just disabled the test currently

.github/workflows/swift.yaml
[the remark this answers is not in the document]
jamieQ: seems like it adds a couple additional minutes to the CI time for that job. if it ends up being an issue i think we can probably drop it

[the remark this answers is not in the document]
watt: > what things have you seen break?

Apple changes something on their backend and it just breaks this completely, blocking PRs in the meantime. The last time it happened we ended up disabling the check in Blueprint because it took more than several days to resolve.

PR #348, opened 2025-05-20
Fix deprecation warning

gilgtm:
Xcode was offering to update my deprecated method call but the signature it chose was incorrect

## Checklist

- [x] Unit Tests
- [x] UI Tests


mjohnson12 (approved):

jamieQ (approved):
good catch – thanks!

PR #349, opened 2025-05-20
[BREAKING CHANGE]: update WorkflowAction API to expose access to Workflow instance values

jamieQ:
this BREAKING change updates the WorkflowAction API to pass through a new parameter in the `apply()` method that allows reading values off the corresponding node's current Workflow instance. the new method signature is:

```swift
func apply(toState state: inout State, context: ApplyContext<WorkflowType>) -> Output?
```

which mirrors the `render()` API, but uses a new `ApplyContext` parameter to expose access to the runtime-managed data. the public API for `ApplyContext` exposes a read-only subscript for accessing Workflow values via keypath:

```swift
public struct ApplyContext<WorkflowType: Workflow> {
    public subscript<Value>(
        workflowValue keyPath: KeyPath<WorkflowType, Value>
    ) -> Value 
}
```
---

for reviewing purposes, i'd suggest using the 'commit filter' function to exclude the commit labeled 'AUTOMATED' as that one is essentially just noise where a bunch of new parameters were added.

also, there's definitely some more documentation to be written here, but i've currently punted on that somewhat. do feel free to point out places you'd like to see anything specific though.

### DIFF
--- a/Workflow/Sources/ApplyContext.swift
+++ b/Workflow/Sources/ApplyContext.swift
@@ -24,8 +24,39 @@ protocol ApplyContextType<WorkflowType> {
 }
 
 /// Runtime context passed as a parameter to `WorkflowAction`'s `apply()` method
-/// that provides an integration point with the runtime that can be used to read property values
-/// off of the current `Workflow` instance.
+/// that provides an integration point with the runtime that can be used to read values from
+/// the current `Workflow` instance.
+///
+/// Read-only access to `Workflow` values is exposed via the `subscript[workflowValue:]` API,
+/// which accepts a read-only `KeyPath` to a `Workflow`'s value.
+///
+/// Usage example:
+///
+/// ```swift
+///     struct MyWorkflow: Workflow {
+///         let shouldSuppressOutput: Bool
+///
+///         // ... implementation ...
+///     }
+///
+///     enum MyAction: WorkflowAction {
+///         typealias WorkflowType = MyWorkflow
+///
+///         case one
+///         case two
+///
+///         func apply(toState state: inout WorkflowType.State, context: ApplyContext<WorkflowType>) -> WorkflowType.Output? {
+///             // Make conditional choices based on the `Workflow`'s instance values
+///             let shouldSuppressOutput = context[workflowValue: \.shouldSuppressOutput]
+///             if shouldSuppressOutput { return nil }
+///
+///             // ... implementation ...
+///         }
+///     }
+/// ```
+///
+/// > Warning: The instance of this type passed to the `apply()` method should not escape from that method.
+/// Attempting to access the instance after the `apply()` method has returned is a client error and will crash.
 public struct ApplyContext<WorkflowType: Workflow> {
     let wrappedContext: any ApplyContextType<WorkflowType>
 
--- a/Workflow/Sources/ApplyContext.swift
+++ b/Workflow/Sources/ApplyContext.swift
@@ -1,0 +1,87 @@
+/*
+ * Copyright Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// Internal utility protocol so that `WorkflowTesting` can provide an alternate implementation
+protocol ApplyContextType<WorkflowType> {
+    associatedtype WorkflowType: Workflow
+
+    subscript<Value>(
+        workflowValue keyPath: KeyPath<WorkflowType, Value>
+    ) -> Value { get }
+}
+
+/// Runtime context passed as a parameter to `WorkflowAction`'s `apply()` method
+/// that provides an integration point with the runtime that can be used to read property values
+/// off of the current `Workflow` instance.
+public struct ApplyContext<WorkflowType: Workflow> {
+    let wrappedContext: any ApplyContextType<WorkflowType>
+
+    init<Impl: ApplyContextType>(_ context: Impl)
+        where Impl.WorkflowType == WorkflowType
+    {
+        self.wrappedContext = context
+    }
+
+    public subscript<Value>(
+        workflowValue keyPath: KeyPath<WorkflowType, Value>
+    ) -> Value {
+        wrappedContext[workflowValue: keyPath]
+    }
+}
+
+extension ApplyContext {
+    static func make<Wrapped: ApplyContextType>(
+        implementation: Wrapped
+    ) -> ApplyContext<Wrapped.WorkflowType>
+        where Wrapped.WorkflowType == WorkflowType
+    {
+        ApplyContext(implementation)
+    }
+}
+
+// FIXME: this is currently a class so that we can zero out the storage
+// when it's invalidated. it'd be nice to eventually make the `ApplyContext`
+// type itself `~Escapable` since that's really the behavior that we want
+// to enforce.
+
+/// The `ApplyContext` used by the Workflow runtime when applying actions.
+final class ConcreteApplyContext<WorkflowType: Workflow>: ApplyContextType {
+    private(set) var storage: WorkflowType?
+
+    private var validatedStorage: WorkflowType {
+        guard let storage else {
+            fatalError("Attempt to use an ApplyContext for Workflow of type '\(WorkflowType.self)' after it was invalidated. The context is only valid during a call to an apply(...) method.")
+        }
+
+        return storage
+    }
+
+    init(storage: WorkflowType) {
+        self.storage = storage
+    }
+
+    subscript<Value>(
+        workflowValue keyPath: KeyPath<WorkflowType, Value>
+    ) -> Value {
+        validatedStorage[keyPath: keyPath]
+    }
+
+    // MARK: -
+
+    func invalidate() {
+        storage = nil
+    }
+}
--- a/Workflow/Sources/RenderContext.swift
+++ b/Workflow/Sources/RenderContext.swift
@@ -184,7 +184,7 @@ extension RenderContext {
     ) -> Sink<Event> {
         makeSink(of: AnyWorkflowAction.self)
             .contraMap { event in
-                AnyWorkflowAction<WorkflowType> { state in
+                AnyWorkflowAction<WorkflowType> { state, _ in
                     onEvent(event, &state)
                 }
             }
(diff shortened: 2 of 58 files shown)


Workflow/Sources/RenderContext.swift
```diff
@@ -180,12 +180,17 @@ protocol RenderContextType: AnyObject {
 extension RenderContext {
     public func makeSink<Event>(
         of eventType: Event.Type,
-        onEvent: @escaping (Event, inout WorkflowType.State) -> WorkflowType.Output?
+        onEvent: @escaping (
```
jamieQ, line 183: - [ ] TODO: resolve this

jamieQ: just reverted it for now. we can make an additive change if we need to in the future

Workflow/Sources/WorkflowAction.swift
```diff
@@ -71,16 +81,35 @@ public struct AnyWorkflowAction<WorkflowType: Workflow>: WorkflowAction {
...
+        fileID: StaticString = #fileID,
+        line: UInt = #line
+    ) {
+        self.init(
+            { state, _ in apply(&state) },
+            fileID: fileID,
+            line: line
+        )
+    }
+
```
blakemcanally, line 99: This means we don't have to go update a bunch of existing callsites that use the current `apply` API right?

Workflow/Sources/WorkflowAction.swift
```diff
@@ -27,14 +27,22 @@ public protocol WorkflowAction<WorkflowType> {
     ///
     /// - Returns: An optional output event for the workflow. If an output event is returned, it will be passed up
     ///            the workflow hierarchy to this workflow's parent.
-    func apply(toState state: inout WorkflowType.State) -> WorkflowType.Output?
+    func apply(
```
g-mark, line 30: Any thoughts around offering both the old and the new functions to avoid changes to lots of files?

I.e., if you added the new one, along with a default impl that called the old one, could we get away with only changing existing action types that need access to the `context`?

Workflow/Sources/WorkflowAction.swift
```diff
@@ -27,14 +27,22 @@ public protocol WorkflowAction<WorkflowType> {
...
+        toState state: inout WorkflowType.State,
+        context: ApplyContext<WorkflowType>
+    ) -> WorkflowType.Output?
+}
```
robmaceachern, line 34: I was wondering if there's some clever way to introduce this in a non-breaking way with protocol extensions (to ease the initial integration/migration) but I don't think there is... Throwing it out there in case someone smarter than me sees a way to make it work.

Workflow/Sources/WorkflowAction.swift
jamieQ: this was [explored](https://github.com/square/workflow-swift/pull/259) a few times in the past and i believe the conclusion was that it would be better to 'rip the bandaid off' so to speak. IMO the primary problem with trying to be source compatible is that you end up having to expose sort of redundant protocol requirements. yes, you can default the one with the new param to forward to the one without it, but it's still sort of an odd API. i think ideally you'd want them to be mutually exclusive, but i'm not sure that's expressible.

jamieQ: that being said – if you've got alternate ideas, happy to hear them!

kyleve: Jamie has already explored ways to automate a large part of this migration, so I am fine with making the breaking change. It's why we have semver!

g-mark: Yeah, that makes sense - thanks for sharing the previous exploration!

mjohnson12 (approved):
Much better than having to store stuff in State that is not state.  👍

Workflow/Sources/WorkflowAction.swift
square-tomb: I tried that out [here](https://github.com/square/workflow-swift/pull/259/files#diff-9ec63dace15e874dc95af768751e6b3fb033fe72876b3916e7b86ea2da80cdafR39-R43), and it works, but as @jamieQ mentioned [above](https://github.com/square/workflow-swift/pull/349#discussion_r2100851792) requires confusingly exposing a second "core" protocol for actions.

Workflow/Sources/ApplyContext.swift
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// Internal utility protocol so that `WorkflowTesting` can provide an alternate implementation
+protocol ApplyContextType<WorkflowType> {
```
jamieQ, line 18: i'm no longer certain this abstraction is needed. i mostly copied it from the analogous thing done for RenderTester, but unlike in that case i don't think the testing infra relies on 'real' internals to apply actions.

Workflow/Sources/WorkflowAction.swift
```diff
@@ -27,14 +27,22 @@ public protocol WorkflowAction<WorkflowType> {
     ///
```
watt, line 27: nit: document new parameter

Workflow/Sources/StateMutationSink.swift
```diff
@@ -44,7 +44,7 @@ public struct StateMutationSink<WorkflowType: Workflow> {
     ///   - update: The `State` mutation to perform.
     public func send(_ update: @escaping (inout WorkflowType.State) -> Void) {
```
watt, line 45: thoughts on exposing the ApplyContext to StateMutationSink?

Workflow/Sources/ApplyContext.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+// FIXME: this is currently a class so that we can zero out the storage
+// when it's invalidated. it'd be nice to eventually make the `ApplyContext`
+// type itself `~Escapable` since that's really the behavior that we want
+// to enforce.
+
+/// The `ApplyContext` used by the Workflow runtime when applying actions.
+final class ConcreteApplyContext<WorkflowType: Workflow>: ApplyContextType {
```
watt, line 61: This type stores a workflow, it _is_ the storage.
```suggestion
final class ContextStorage<WorkflowType: Workflow>: ApplyContextType {
```

Workflow/Sources/ApplyContext.swift
```diff
@@ -0,0 +1,87 @@
...
+    let wrappedContext: any ApplyContextType<WorkflowType>
+
+    init<Impl: ApplyContextType>(_ context: Impl)
+        where Impl.WorkflowType == WorkflowType
+    {
+        self.wrappedContext = context
+    }
+
+    public subscript<Value>(
+        workflowValue keyPath: KeyPath<WorkflowType, Value>
```
watt, line 39: Decided not to use dynamic property lookup?

Workflow/Sources/ApplyContext.swift
```diff
@@ -0,0 +1,87 @@
...
+public struct ApplyContext<WorkflowType: Workflow> {
+    let wrappedContext: any ApplyContextType<WorkflowType>
+
+    init<Impl: ApplyContextType>(_ context: Impl)
+        where Impl.WorkflowType == WorkflowType
+    {
+        self.wrappedContext = context
+    }
+
+    public subscript<Value>(
```
watt, line 38: This is missing docs, but I think it would actually be more effective to document usage on the `ApplyContext` type, with a small example of how to read workflow values.

Workflow/Sources/WorkflowNode.swift
```diff
@@ -218,7 +218,14 @@ extension WorkflowNode {
         defer { observerCompletion?(state, output) }
 
         /// Apply the action to the current state
-        output = action.apply(toState: &state)
+        do {
+            // TODO: can we avoid instantiating a class here somehow?
```
watt, line 222: Is this TODO something you expect to address before merging?

Workflow/Tests/ApplyContextTests.swift
```diff
@@ -0,0 +1,79 @@
...
+
+@testable import Workflow
+
+@MainActor
+struct ApplyContextTests {
+    @Test
+    func concreteApplyContextInvalidatedAfterUse() async throws {
+        var escapedContext: ApplyContext<EscapingContextWorkflow>?
+        let onApply = { (context: ApplyContext<EscapingContextWorkflow>) in
+            #expect(context[workflowValue: \.property] == 42)
```
watt, line 27: These are the first `#expect`-based tests in Workflow, I believe. Just to confirm, did you check that failures are caught in CI?

WorkflowTesting/Sources/WorkflowActionTester.swift
```diff
@@ -163,3 +184,34 @@ extension WorkflowActionTester where WorkflowType.Output: Equatable {
         }
     }
 }
+
+// MARK: - ApplyContext
+
+struct TestApplyContext<Wrapped: Workflow>: ApplyContextType {
+    enum TestContextKind {
+        case workflow(Wrapped)
+        // TODO: flesh this out to support 'just in time' values
```
watt, line 193: and this TODO?

watt (approved):
I've got several questions but no big blockers. Thanks for tackling this!

I also noticed you're adding several TODO comments, where it's not entirely clear if you were intending to address them with this PR or if you're just marking new tech debt.

Workflow/Tests/ApplyContextTests.swift
jamieQ: thanks for the callout. yes, i did a 'canary' run [yesterday](https://github.com/square/workflow-swift/pull/350) with a test that should have failed CI, and it seemed to behave as expected

Workflow/Sources/StateMutationSink.swift
jamieQ: i think that probably makes sense to do, but would prefer to tackle as a follow-up for scope-management reasons, unless you feel this will immediately be desired. i _think_ it's a straightforward plumbing change, and i may have had it implemented at one point but removed it.

Workflow/Sources/ApplyContext.swift
jamieQ: yes, for now anyway. this will make callsites more verbose, but also easier to locate in the source code, and somewhat more explicit about what's going on.

Workflow/Sources/WorkflowNode.swift
jamieQ: not really, more of a self-note. i'll update these to `FIXME` to make it more clear they are potential outstanding improvements, and not required pending tasks

WorkflowTesting/Sources/WorkflowActionTester.swift
jamieQ: this will be a follow-up

Workflow/Sources/ApplyContext.swift
jamieQ: added some more context & a usage example to the doc comment on the containing type

square-tomb (approved):

PR #353, opened 2025-06-02
Move AnyAsyncSequenceWorker into WorkflowConcurrency as AsyncSequenceWorker

mjohnson12:
This moves `AnyAsyncSequenceWorker` from `WorkflowExperimental` in register into `WorkflowConcurrency` as `AsyncSequenceWorker`.

** Update **
After pairing with @jamieQ I've pushed out an update that includes his change that adds a `Sequence` associated type that can be used to require `Element` from `AsyncSequence` to match `Output`.
This allows the worker workflow to be able to get typed output from the `for await`.

## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
+++ b/WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
@@ -9,16 +9,16 @@ import Workflow
 /// If there is, and if the workers are 'equivalent', the context leaves the existing worker running.
 ///
 /// If there is not an existing worker of this type, the context will kick off the new worker (via `run`).
-public protocol AsyncSequenceWorker<Output>: AnyWorkflowConvertible where Rendering == Void {
+public protocol AsyncSequenceWorker<Output>: AnyWorkflowConvertible where Rendering == Void, Output: Sendable {
     /// The type of output events returned by this worker.
-    associatedtype Output
+    associatedtype Sequence: AsyncSequence where Sequence.Element == Output
 
     // In iOS 18+ we can do:
     // func run() -> any AsyncSequence<Output, Never>
-    // And then remove the casting in the side effect
 
     /// Returns an `AsyncSequence` to execute the work represented by this worker.
-    func run() -> any AsyncSequence
+    func run() -> Sequence
+
     /// Returns `true` if the other worker should be considered equivalent to `self`. Equivalence should take into
     /// account whatever data is meaningful to the task. For example, a worker that loads a user account from a server
     /// would not be equivalent to another worker with a different user ID.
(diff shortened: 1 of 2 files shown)


WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
```diff
@@ -0,0 +1,72 @@
...
+    func render(state: State, context: RenderContext<AsyncSequenceWorkerWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: state) { lifetime in
+            let task = Task {
+                for try await output in worker.run() {
+                    // Not necessary in iOS 18+ once we can use AsyncSequence<Output, Never>
+                    guard let output = output as? Output else {
+                        fatalError("Unexpected output type \(type(of: output)) from worker \(worker)")
+                    }
+                    await sendAction(output: output, sink: sink)
```
mjohnson12, line 58: If I change this to do what the async function worker does to call send on `sink` by adding:

`let send: @MainActor (Output) -> Void = sink.send`

It crashes the compiler

amorde (approved):

WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    func render(state: State, context: RenderContext<AsyncSequenceWorkerWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: state) { lifetime in
+            let task = Task {
+                for try await output in worker.run() {
+                    // Not necessary in iOS 18+ once we can use AsyncSequence<Output, Never>
+                    guard let output = output as? Output else {
```
jamieQ, line 55: hmm... can we not constrain the type in a better manner than this? when did `AsyncSequence` get primary associated types?

jamieQ: i found [this idea](https://forums.swift.org/t/primary-associated-types-for-asyncsequences/67099/2) – would something like that give us more expressivity? maybe it could be conditional based on deployment target?

WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
```diff
@@ -0,0 +1,69 @@
...
+        if !worker.isEquivalent(to: previousWorkflow.worker) {
+            state = UUID()
+        }
+    }
+
+    func render(state: State, context: RenderContext<AsyncSequenceWorkerWorkflow>) -> Rendering {
+        let sink = context.makeSink(of: AnyWorkflowAction.self)
+        context.runSideEffect(key: state) { lifetime in
+            let task = Task {
+                for try await output in worker.run() {
```
jamieQ, line 53: is this actually throwing? if so, where is the error handled?

WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import Workflow
```
jamieQ, line 2: have you tried silencing the concurrency warnings via use of Sendable requirements/annotations, or other means? since this is new, should we enforce that now (even though it's inconsistent with other API)?

WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
mjohnson12: In iOS 18 it works great we can do `func run() -> any AsyncSequence<Output, Never>` but since we can't use a minimum of iOS 18 we are currently stuck with `any AsyncSequence`

WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
mjohnson12: I did try `adding `@preconcurrency` like it suggested and it still gave the same warnings.
Could we enforce `Sendable` on `Output` and it still be convertible to AnyWorkflow since in Workflow Output is not Sendable?

WorkflowConcurrency/Sources/AsyncSequenceWorker.swift
mjohnson12: Sorry your second comment didn't show for me in Github until now...Will consumers be creating their own custom types that implement `AsyncSequence` or just using `AsyncStream` and swift async algos?

PR #354, opened 2025-06-24
[feat]: introduce runtime configuration SPI

jamieQ:
introduces the notion of a 'runtime configuration' in order to support opting into new behaviors to the Workflow runtime over time. currently this is behind an SPI, and should be considered experimental and exempt from semantic versioning standards.

### DIFF
--- a/Workflow/Sources/RuntimeConfiguration.swift
+++ b/Workflow/Sources/RuntimeConfiguration.swift
@@ -17,7 +17,7 @@
 /// System for managing configuration options for Workflow runtime behaviors.
 /// - important: These interfaces are subject to breaking changes without corresponding semantic
 /// versioning changes.
-@_spi(RuntimeConfig)
+@_spi(WorkflowRuntimeConfig)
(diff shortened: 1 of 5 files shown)


amorde (approved):

Workflow/Sources/RuntimeConfiguration.swift
```diff
@@ -0,0 +1,102 @@
...
+    static var _currentConfiguration: Configuration?
+
+    static var _bootstrapConfiguration = BootstrappableConfiguration()
+
+    /// Bootstrap the workflow runtime with the given configuration.
+    /// This can only be called once per process and must be called from the main thread.
+    ///
+    /// - Parameter configuration: The runtime configuration to use.
+    @MainActor
+    public static func bootstrap(
```
watt, line 32: not at all a blocker, but IMO the one-time bootstrapping requirement is overkill vs having a static var that can be changed at any time

watt (approved):

PR #370, opened 2025-08-19
Updated build to Xcode 16.5 and iOS 18.5

mjohnson12:
## Checklist

- [ ] Unit Tests
- [ ] UI Tests

### DIFF
--- a/.github/workflows/swift.yaml
+++ b/.github/workflows/swift.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 env:
-  XCODE_VERSION: 16.1.0
+  XCODE_VERSION: 16.4.0
   TUIST_TEST_DEVICE: iPad (10th generation)
   TUIST_TEST_PLATFORM: iOS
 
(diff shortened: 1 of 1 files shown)


.github/workflows/swift.yaml
```diff
@@ -85,13 +85,13 @@ jobs:
...
   package-tests:
-    runs-on: macos-latest
+    runs-on: macos-14
 
     steps:
     - uses: actions/checkout@v4
 
     - name: Switch Xcode
-      run: sudo xcode-select -s /Applications/Xcode_${{ env.XCODE_VERSION }}.app
+      run: sudo xcode-select -s /Applications/Xcode_16.1.app
```
jamieQ, line 94: why are we pinning this for now again? some test fails on newer macos?

mjohnson12: The image for MacOS 14 doesn't have Xcode 16.4 so that's why I had to hard code this.

PR #377, opened 2025-09-03
chore: Bump Perception to 2.0.6

robmaceachern:
## Checklist

**Bump perception to 2.0.6**
**`tuist install --path Samples` to update Sample resolved dependencies**

The Perception version specified in Package.swift is now `1.5.0"..<"3.0.0`. This allows easier upgrade and downgrades by consumers.

I've opened this in favor of https://github.com/square/workflow-swift/pull/376 because Perception has been [updated](https://github.com/pointfreeco/swift-perception/pull/156) following Apple's reversion of https://github.com/swiftlang/swift/pull/83436 and it avoids test updates on our end for now.

I didn't run the `tuist install` with the `--update` flag this time either. There doesn't seem to be a way to scope the updates to only Perception and it's sub-dependencies and I didn't want all the other clutter associated with this change.

- [x] Unit Tests
- [x] UI Tests
- [x] Resolve new Perception warning in ios-register: https://github.com/squareup/ios-register/pull/131811
- [x] Resolve new instances of Perception warnings in Market: https://github.com/squareup/market/pull/11042

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -59,7 +59,7 @@ let package = Package(
         .package(url: "https://github.com/pointfreeco/swift-case-paths", from: "1.5.5"),
         .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "1.0.0"),
         .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.4.0"),
-        .package(url: "https://github.com/pointfreeco/swift-perception", from: "2.0.0"),
+        .package(url: "https://github.com/pointfreeco/swift-perception", "1.5.0"..<"3.0.0"),
         .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.3.3"),
         .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.0.0"),
     ],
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -59,7 +59,7 @@ let package = Package(
         .package(url: "https://github.com/pointfreeco/swift-case-paths", from: "1.5.5"),
         .package(url: "https://github.com/pointfreeco/swift-identified-collections", from: "1.0.0"),
         .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.4.0"),
-        .package(url: "https://github.com/pointfreeco/swift-perception", from: "1.5.0"),
+        .package(url: "https://github.com/pointfreeco/swift-perception", from: "2.0.0"),
```
watt, line 62: Since we didn't have to make any code changes, what do you think about allowing a range here, like we do for swift-syntax? Then we could freely upgrade or rollback downstream without having to touch workflow again.

something like this?
```suggestion
        .package(url: "https://github.com/pointfreeco/swift-perception", "1.5.0"..<"3.0.0"),
```

robmaceachern: Yeah I think that makes sense.

I think the only risk is we start depending on newer API and don't remember to bump it back up. Not a big deal, it'd just be like https://github.com/square/workflow-swift/pull/375

jamieQ (approved):

watt (approved):

PR #379, opened 2025-09-16
[DNM][fix]: refactor deferred action handling to sidestep reentrancy issue

jamieQ:
this is a draft of small change we could make to automatically enqueue event handling code that is triggered while a node is already handling an event. i don't think this will handle all cases, but it is a pretty minor change that does seem to resolve at least one that appears to sometimes surface in client code. a more holistic approach is explored in https://github.com/square/workflow-swift/pull/381

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -15,6 +15,7 @@
  */
 
 import Dispatch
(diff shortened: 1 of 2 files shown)


mjohnson12 (approved):
Yeah I agree this looks pretty safe to add

PR #380, opened 2025-09-17
[refactor]: improve diagnostics for escaped context instances

jamieQ:
### Issue

the `{Render|Apply}Context` types should not escape from their respective methods, as this can allow library internals to 'leak out' and cause various issues[^1]. primarily this can result in lifetime discrepancies or surprising relationships in the object graph. historically we've enforced this invariant 'lazily' by setting a bit on the instances and checking that they are still 'valid' during access through their public API. however, it recently occurred to me that we could probably be a bit more proactive about this, and perform a unique reference check immediately after returning from the client callouts.

### Description

- adds logic to detect escaping context instances (debug mode only)
- adds the 'IssueReporting' dependency to Workflow so we can better test the logic in our unit tests
- minor incidental refactoring of the Package file
- updates unit tests appropriately

[^1]: this is really more of an issue for RenderContext, as it doesn't yet zero-out its internal storage when invalidated. since ApplyContext is newer, we didn't make that mistake twice.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -61,14 +61,18 @@ let package = Package(
         .package(url: "https://github.com/pointfreeco/swift-macro-testing", from: "0.4.0"),
         .package(url: "https://github.com/pointfreeco/swift-perception", "1.5.0" ..< "3.0.0"),
         .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.3.3"),
-        .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.0.0"),
+        // 'xctest-dynamic-overlay' is actually for "IssueReporting"
+        .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.2.2"),
     ],
     targets: [
         // MARK: Workflow
 
         .target(
             name: "Workflow",
-            dependencies: ["ReactiveSwift"],
+            dependencies: [
+                .product(name: "IssueReporting", package: "xctest-dynamic-overlay"),
+                .product(name: "ReactiveSwift", package: "ReactiveSwift"),
+            ],
             path: "Workflow/Sources"
         ),
         .target(
@@ -76,6 +80,7 @@ let package = Package(
             dependencies: [
                 "Workflow",
                 .product(name: "CustomDump", package: "swift-custom-dump"),
+                .product(name: "IssueReporting", package: "xctest-dynamic-overlay"),
             ],
             path: "WorkflowTesting/Sources",
             linkerSettings: [.linkedFramework("XCTest")]
@@ -125,7 +130,10 @@ let package = Package(
 
         .target(
             name: "WorkflowReactiveSwift",
-            dependencies: ["ReactiveSwift", "Workflow"],
+            dependencies: [
+                "Workflow",
+                .product(name: "ReactiveSwift", package: "ReactiveSwift"),
+            ],
             path: "WorkflowReactiveSwift/Sources"
         ),
         .target(
@@ -139,7 +147,10 @@ let package = Package(
 
         .target(
             name: "WorkflowRxSwift",
-            dependencies: ["RxSwift", "Workflow"],
+            dependencies: [
+                "Workflow",
+                .product(name: "RxSwift", package: "RxSwift"),
+            ],
             path: "WorkflowRxSwift/Sources"
         ),
         .target(
--- a/Workflow/Sources/Debugging.swift
+++ b/Workflow/Sources/Debugging.swift
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+#if DEBUG
+import IssueReporting
+#endif
+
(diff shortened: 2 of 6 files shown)


Package.swift
```diff
@@ -61,21 +61,26 @@ let package = Package(
...
         .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.0.0"),
     ],
     targets: [
         // MARK: Workflow
 
         .target(
             name: "Workflow",
-            dependencies: ["ReactiveSwift"],
+            dependencies: [
+                .product(name: "IssueReporting", package: "xctest-dynamic-overlay"),
```
jamieQ, line 73: with this change i did start getting some weird Tuist warnings i don't really understand about statically linking things multiple times... didn't _seem_ to cause any build issues, but not sure if i'm doing something wrong.

Workflow/Sources/Debugging.swift
```diff
@@ -168,3 +168,24 @@ extension WorkflowUpdateDebugInfo {
...
+import IssueReporting
+
+func diagnoseEscapedReference(
+    to object: consuming some AnyObject,
+    _ message: @autoclosure () -> (ObjectIdentifier) -> String
+) {
+    var maybeUniqueReference = consume object
+    if !isKnownUniquelyReferenced(&maybeUniqueReference) {
+        if let _ = TestContext.current {
+            reportIssue(message()(ObjectIdentifier(maybeUniqueReference)))
```
jamieQ: - [x] TODO: figure out if autoclosures make any sense here or not

Package.swift
```diff
@@ -125,7 +130,10 @@ let package = Package(
 
         .target(
             name: "WorkflowReactiveSwift",
-            dependencies: ["ReactiveSwift", "Workflow"],
+            dependencies: [
+                "Workflow",
+                .product(name: "ReactiveSwift", package: "ReactiveSwift"),
```
jamieQ, line 135: standardized some of these so now all the 'package-level' dependencies are just strings, and the external ones are `.product()`

Workflow/Sources/Debugging.swift
jamieQ: decided they do not

Package.swift
jamieQ: they are like this:

```bash
The following warnings need attention:
 · Target 'IssueReportingPackageSupport' has been linked from target 'Workflow' and target 'WorkflowRxSwift-Tests', it is a static product so may introduce unwanted side effects.
 · Target 'IssueReportingPackageSupport' has been linked from target 'Workflow' and target 'WorkflowTesting-Tests', it is a static product so may introduce unwanted side effects.
 · Target 'IssueReporting' has been linked from target 'Workflow' and target 'WorkflowConcurrencyTesting-Tests', it is a static product so may introduce unwanted side effects.
// etc
 ```

Comment (jamieQ) [reply Package.swift:73]: seems this may be an artifact of the version of Tuist we're on – will look into updating that separately, and ignore this for now

**Review by amorde — APPROVED** [review]
Very cool! Only question is if there's any concerns having the `Workflow` target now depend on the `IssueReporting` framework from the xctest overlay. I suppose it should be fine given we only use it in `#if DEBUG`, but perhaps there's some potential dependency conflicts for downstream consumers using SPM if they use a different version or something like that.

**Comment by jamieQ** [comment]
> Only question is if there's any concerns having the `Workflow` target now depend on the `IssueReporting` framework from the xctest overlay. I suppose it should be fine given we only use it in `#if DEBUG`, but perhaps there's some potential dependency conflicts for downstream consumers using SPM if they use a different version or something like that.

yeah, i guess i'm unsure exactly what the theoretical risk is here... we have had the dependency in the package for a while, but i guess not as part of the Workflow target, so maybe that could cause problems? i think we're as permissive with the version as possible though, so maybe it's okay.

i guess i should probably also look into if the library gets built & linked in non-debug builds even if it's only referenced behind conditional compilation directives...

---

edit: i think since we've had the implicit dependency already we're okay and won't cause new issues for people due to dependency resolution (i set the dependency to what seemed like the actual minimum version we require `1.2.2`). also seems there's not really a way to do non-debug build & link only, but also i think this is probably okay since it's effectively only used in debug and nothing is publicly re-exported.

File: Workflow/Tests/ApplyContextTests.swift [hunk 4]
```diff
@@ -14,18 +14,18 @@
...
+import IssueReporting
+import XCTest
 
 @testable import Workflow
 
 @MainActor
-struct ApplyContextTests {
-    @Test
-    func concreteApplyContextInvalidatedAfterUse() async throws {
+final class ApplyContextTests: XCTestCase {
```
jamieQ, line 23: converted this to XCTest b/c when testing against the oldest version of the IssueReporting dependency we support (1.2.2), for some reason the test seemed to behave differently and didn't appear to actually 'escape' the context. i assume that means there's a bug somewhere, but did not investigate.

PR #381, opened 2025-09-18
[fix]: add host-level logic for reentrant event handling

jamieQ:
#### Terminology

1. `Sink`: public action handler type vended out to client code. these are the primary channel through which events from the 'outside world' are sent into a Workflow runtime. is a value type that wraps a (weak) reference to internal event handling infrastructure.
1. `ReusableSink`: internal action handler type that receives actions forwarded through from `Sink`s. it is a reference type that is weakly referenced by `Sink`s.
1. `EventPipe`: internal action handler type that implements an event handling state machine. it is a reference type that is owned by either a `ReusableSink` to propagate 'outside' events, or by a `SubtreeManager` to propagate outputs from child workflows.
1. `SubtreeManager`: type that coordinates interactions between a workflow node, its children, and the 'outside world'.

#### Background

the current event handling system in Workflow tracks event processing state locally – each node has a 'subtree manager' which is responsible for orchestrating interactions with both child workflow outputs and events from the 'outside world' sent to the corresponding node. each `SubtreeManager` coordinates a collection of 'event pipes' which handle event propagation; the 'outside world' has (indirect) access to them via 'sinks', and child workflow's communicate 'up' to their parents through `EventPipe` instances. these event pipes also encode the state machine transitions for event handling – if they're sent events in an invalid state they will generally halt the program.

the `EventPipe` state machine consists of the following states:

1. `preparing`: the pipe has been created, but this state indicates a call to the corresponding node's `render()` method is still ongoing.
1. `pending`: the corresponding node's `render()` method is complete, but the full 'render pass' over the entire tree may not yet be.
1. `enabled`: the tree's render pass is finished, and the event pipe is valid to use. this state has a corresponding event handler closure that can be invoked when handling an event.
1. `invalid`: the event pipe is no longer valid to use. a node's event pipes are invalidated as one of the first parts of a render pass.

the currently expected state transitions are:

1. <none> -> `preparing`
    - `preparing` is the initial state
1. `preparing` -> `pending`
    - after a node finishes `render()`
1. `pending` -> `enabled`
    - after the tree finishes a render pass
1. `enabled` -> `invalid`
    - when a node is about to be re-rendered[^1]

[^1]: as an aside, it's reasonable to wonder if we may be allowing nodes to 'linger on' without invalidating their event handling infrastructure appropriately. what happens if a node is rendered in one pass and not rendered in the next? i think in most cases it just deallocates and so implicitly ends up ignoring any subsequent events, but... would probably be good to verify that and formalize what should be happening...

the way the current event pipe state machine works is:

1. during a render pass, every node gets new event pipes created, initially in the `preparing` state. any existing event pipes are set to the `invalid` state.
1. after a node is rendered, the event pipes that were created for that pass are moved to the `pending` state.
1. after a render pass is 'complete' and any Output and a new Rendering have been emitted, the nodes in the tree are walked and all event pipes are moved to the `enabled` state and hooked up with the right callbacks to invoke when they're sent events.

some additional notes about the current implementation:

1. a number of invalid state transitions are banned outright and will cause a runtime trap if attempted.
1. the `ReusableSink` type contains logic to attempt to detect certain forms of reentrancy. specifically it will check if the event pipe state is `pending` and if it is, will enqueue the forwarding of the event into the future.
1. there is some limited reentrancy detection implemented when event pipes in the `invalid` state are reentrantly messaged.

#### Issue

the existing implementation seems to have generally worked reasonably well in practice, but there are cases when it falls down. in particular, two problematic cases that have been seen 'in the wild' are:

1. reentrant action emissions from synchronous side effects. perhaps the 'canonical' example of this is when an action is being processed that, during processing, leads to a change in `UIResponder` state (.e.g `resignFirstResponder()` is called), and some _other_ part of the Workflow tree responds to that change by emitting another action.

1. APIs that do manual `RunLoop` spinning, leading to reentrant action handling. one instance of this we've seen is when the UI layer responds to a new rendering by deriving an attributed string from HTML via some the `UIKit` extension methods it can end up spinning the main thread's run loop waiting for `WebKit` to do HTML processing, and if there are other Workflow events that have been enqueued they will cause the runtime to attempt to reentrantly process an event (generally leading to a `fatalError()`).

as the existing implementation models the event handling state machine at the node level, it seems ill-equipped to deal with this problem holistically since the 'is the tree currently processing an event' bit of information is inherently a tree-level fact. we could try to augment the existing code with a state that represents 'the runtime is currently handling an event', but that seems somewhat awkward to fit into the existing model since a node would have to walk to the root and then update the whole tree with this new state[^2].

[^2]: except maybe _not_ the path from node to root? see, it seems kind of awkward...

#### Proposed Solution

in this PR, the approach taken to solve this is:

- introduce a new `SinkEventHandler` type to track tree-level event processing state
- plumb a new callback closure through the tree down to the `ReusableSink` instances (which are responsible for forwarding 'external' events into the `EventPipe` machinery)
    - the new callback takes two parameters: a closure to be immediately invoked if there is no event currently being handled, and a closure to be enqueued and called in the future if there is.
- the callback implementation checks the current state and either invokes the 'run now' closure (adjusting processing state as appropriate) or enqueues the 'run later' closure.

#### Alternatives

i also drafted a more minimal change to address the new test case that i added which simulates the 'spinning the run loop during a render pass' problem in https://github.com/square/workflow-swift/pull/379. instead of changing the plumbing and adding tree-level state tracking, it just changes how the enqueing logic in the `ReusableSink` implementation works. previously the enqueuing logic would defer handling and unconditionally forward the event through after the async block was processed. after the change, the method becomes fully recursive, so will check whether the original action should be enqueued again. while this approach requires changing less code, it is also less of a 'real' fix, as it won't solve cases in which someone, say, emits a second sink event targeting an ancestor node in the tree while a first sink event is being handled.

---

#### Updates

after initial feedback, made the following changes:

- made the new event handling behavior conditional and added a `RuntimeConfig` value to opt into it
- added a queue precondition check into the new `ReusableSink` event handling logic
- removed the `initializing` state from `SinkEventHandler` in favor of just 'busy' and 'ready'
- added a mechanism to explicitly enter the `busy` state while invoking a block so that the `WorkflowHost` can update the root node and ensure no event handling can synchronously occur during that process
- added several new test cases (and minor refactoring of existing test utilities)

### DIFF
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -281,7 +281,10 @@ extension WorkflowNode.SubtreeManager {
         func makeSink<Action: WorkflowAction>(
             of actionType: Action.Type
         ) -> Sink<Action> where WorkflowType == Action.WorkflowType {
-            let reusableSink = sinkStore.findOrCreate(actionType: Action.self)
+            let reusableSink = sinkStore.findOrCreate(
+                actionType: Action.self,
+                onSinkEvent: hostContext.onSinkEvent
+            )
 
             let sink = Sink<Action> { [weak reusableSink] action in
                 WorkflowLogger.logSinkEvent(ref: SignpostRef(), action: action)
@@ -320,7 +323,10 @@ extension WorkflowNode.SubtreeManager {
             self.usedSinks = [:]
         }
 
-        mutating func findOrCreate<Action: WorkflowAction>(actionType: Action.Type) -> ReusableSink<Action> {
+        mutating func findOrCreate<Action: WorkflowAction>(
+            actionType: Action.Type,
+            onSinkEvent: @escaping OnSinkEvent
+        ) -> ReusableSink<Action> {
             let key = ObjectIdentifier(actionType)
 
             let reusableSink: ReusableSink<Action>
@@ -334,7 +340,7 @@ extension WorkflowNode.SubtreeManager {
                 reusableSink = usedSink
             } else {
                 // Create a new reusable sink.
-                reusableSink = ReusableSink<Action>()
+                reusableSink = ReusableSink<Action>(onSinkEvent: onSinkEvent)
             }
 
             usedSinks[key] = reusableSink
@@ -345,30 +351,33 @@ extension WorkflowNode.SubtreeManager {
 
     /// Type-erased base class for reusable sinks.
     fileprivate class AnyReusableSink {
+        /// The callback to invoke when an event is to be handled.
+        let onSinkEvent: OnSinkEvent
         var eventPipe: EventPipe
 
-        init() {
+        init(onSinkEvent: @escaping OnSinkEvent) {
+            self.onSinkEvent = onSinkEvent
             self.eventPipe = EventPipe()
         }
     }
 
     fileprivate final class ReusableSink<Action: WorkflowAction>: AnyReusableSink where Action.WorkflowType == WorkflowType {
         func handle(action: Action) {
-            let output = Output.update(
-                action,
-                source: .external,
-                subtreeInvalidated: false // initial state
-            )
+            let perform: () -> Void = {
+                let output = Output.update(
+                    action,
+                    source: .external,
+                    subtreeInvalidated: false // initial state
+                )
 
-            if case .pending = eventPipe.validationState {
-                // Workflow is currently processing an `event`.
-                // Scheduling it to be processed after.
-                DispatchQueue.workflowExecution.async { [weak self] in
-                    self?.eventPipe.handle(event: output)
-                }
-                return
+                self.eventPipe.handle(event: output)
             }
-            eventPipe.handle(event: output)
+
+            let enqueue: () -> Void = { [weak self] in
+                self?.handle(action: action)
+            }
+
+            onSinkEvent(perform, enqueue)
         }
     }
 }
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -325,7 +325,7 @@ extension WorkflowNode.SubtreeManager {
 
         mutating func findOrCreate<Action: WorkflowAction>(
             actionType: Action.Type,
-            onSinkEvent: @escaping OnSinkEvent
+            onSinkEvent: OnSinkEvent?
         ) -> ReusableSink<Action> {
             let key = ObjectIdentifier(actionType)
 
@@ -352,10 +352,10 @@ extension WorkflowNode.SubtreeManager {
     /// Type-erased base class for reusable sinks.
     fileprivate class AnyReusableSink {
         /// The callback to invoke when an event is to be handled.
-        let onSinkEvent: OnSinkEvent
+        let onSinkEvent: OnSinkEvent?
         var eventPipe: EventPipe
 
-        init(onSinkEvent: @escaping OnSinkEvent) {
+        init(onSinkEvent: OnSinkEvent?) {
             self.onSinkEvent = onSinkEvent
             self.eventPipe = EventPipe()
         }
@@ -363,6 +363,36 @@ extension WorkflowNode.SubtreeManager {
 
     fileprivate final class ReusableSink<Action: WorkflowAction>: AnyReusableSink where Action.WorkflowType == WorkflowType {
         func handle(action: Action) {
+            if let onSinkEvent {
+                handleWithSinkEventHandler(action: action, onSinkEvent: onSinkEvent)
+                return
+            }
+
+            // Prior logic
+            let output = Output.update(
+                action,
+                source: .external,
+                subtreeInvalidated: false // initial state
+            )
+
+            if case .pending = eventPipe.validationState {
+                // Workflow is currently processing an `event`.
+                // Scheduling it to be processed after.
+                DispatchQueue.workflowExecution.async { [weak self] in
+                    self?.eventPipe.handle(event: output)
+                }
+                return
+            }
+            eventPipe.handle(event: output)
+        }
+
+        private func handleWithSinkEventHandler(
+            action: Action,
+            onSinkEvent: OnSinkEvent
+        ) {
+            // new `SinkEventHandler` logic
+            dispatchPrecondition(condition: .onQueue(DispatchQueue.workflowExecution))
+
             // If we can process now, forward through the `EventPipe`
             let immediatePerform: () -> Void = {
                 let output = Output.update(
--- a/Workflow/Sources/SubtreeManager.swift
+++ b/Workflow/Sources/SubtreeManager.swift
@@ -363,7 +363,8 @@ extension WorkflowNode.SubtreeManager {
 
     fileprivate final class ReusableSink<Action: WorkflowAction>: AnyReusableSink where Action.WorkflowType == WorkflowType {
         func handle(action: Action) {
-            let perform: () -> Void = {
+            // If we can process now, forward through the `EventPipe`
+            let immediatePerform: () -> Void = {
                 let output = Output.update(
                     action,
                     source: .external,
@@ -373,11 +374,12 @@ extension WorkflowNode.SubtreeManager {
                 self.eventPipe.handle(event: output)
             }
 
-            let enqueue: () -> Void = { [weak self] in
+            // Otherwise, try to recurse again in the future
+            let deferredPerform: () -> Void = { [weak self] in
                 self?.handle(action: action)
             }
 
-            onSinkEvent(perform, enqueue)
+            onSinkEvent(immediatePerform, deferredPerform)
         }
     }
 }
--- a/Workflow/Sources/WorkflowHost.swift
+++ b/Workflow/Sources/WorkflowHost.swift
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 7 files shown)


Workflow/Sources/SubtreeManager.swift
```diff
@@ -345,30 +351,33 @@ extension WorkflowNode.SubtreeManager {
...
+                    subtreeInvalidated: false // initial state
+                )
 
-            if case .pending = eventPipe.validationState {
-                // Workflow is currently processing an `event`.
-                // Scheduling it to be processed after.
-                DispatchQueue.workflowExecution.async { [weak self] in
-                    self?.eventPipe.handle(event: output)
-                }
-                return
```
jamieQ, line 369: trying to think through if we need this check anymore... the `pending` state should only have been possible to be in after a node was rendered during a render pass, but before the rendering & output had finished being emitted. i guess it's maybe conceivable that a child node could do something that would trigger some side effect that sent a sink event to a parent node in the `pending` state... 🤔. i think the 'happy path' case is that it would just be enqueued b/c we'd be re-rendering due to handling an event (and so would hit the 'enqueue' case in the `onSinkEvent` callback), but there is maybe an edge case to consider where the workflow host updates the root node independently (no event being handled) so maybe we should be a bit more cautious here...

mjohnson12 (approved):
Great description and refresher on the types involved.
I did remember how event handling worked but this was great overview.
I like the new SinkEventHandler and OnSinkEvent.
This looks good to me.

Workflow/Sources/SubtreeManager.swift
kcsiegal: I think you've convinced me that it still makes sense to handle this possible edge case, what's the downside of leaving this? Does the dispatch approach not work with the new paradigm?

Workflow/Sources/WorkflowHost.swift
```diff
@@ -176,3 +192,72 @@ extension HostContext {
...
+/// - Parameter perform: The event handler to invoke if the event can be processed immediately.
+/// - Parameter enqueue: The event handler to invoke in the future if the event cannot currently be processed.
+typealias OnSinkEvent = (
+    _ perform: () -> Void,
+    _ enqueue: @escaping () -> Void
+) -> Void
+
+/// Handles events from 'Sinks' such that runtime-level event handling state is appropriately
+/// managed, and attempts to perform reentrant action handling can be detected and dealt with.
+final class SinkEventHandler {
```
kcsiegal, line 218: I find this type's API and behavior very intuitive, I like it!

Workflow/Sources/WorkflowHost.swift
```diff
@@ -176,3 +192,72 @@ extension HostContext {
...
+
+        case .ready:
+            state = .processingEvent
+            defer { state = .ready }
+            perform()
+        }
+    }
+
+    /// Creates the callback that should be invoked by Sinks to handle their event appropriately
+    /// given the `EventHandler`'s current state.
```
kcsiegal: Should this be `SinkEventHandler`?

Workflow/Tests/WorkflowHostTests.swift
```diff
@@ -58,32 +60,84 @@ final class WorkflowHostTests: XCTestCase {
...
-            observedRenderCount += 1
-        }
-        defer { disposable?.dispose() }
-
         // send an event and cause a re-render
         initialRendering.eventHandler()
 
         XCTAssertEqual(observedRenderCount, 1)
+
+        drainMainQueueBySpinningRunLoop()
```
kcsiegal, line 95: Helpful!

Workflow/Sources/WorkflowHost.swift
```diff
@@ -61,6 +64,13 @@ public final class WorkflowHost<WorkflowType: Workflow> {
         observers: [WorkflowObserver] = [],
         debugger: WorkflowDebugger? = nil
     ) {
+        self.eventHandler = SinkEventHandler()
+        assert(
+            eventHandler.state == .initializing,
+            "EventHandler must begin in the `.initializing` state"
+        )
```
nabs-m: Could there ever be situations where right after initializing a `SinkEventHandler`, it wouldn't be in an `initializing` state? I see that this is how you initialize that property below, so I'm wonder if you're trying to protect from future bad changes or if this is some Workflow pattern I'm not familiar with.

kcsiegal (approved):
I am totally new to Workflow's event handling but this seems very reasonable to me, appreciate the detailed context in the PR description and comments.

Workflow/Sources/WorkflowHost.swift
```diff
@@ -176,3 +192,72 @@ extension HostContext {
...
+
+// MARK: - EventHandler
+
+/// Callback signature for the internal `ReusableSink` types to invoke when
+/// they receive an event from the 'outside world'.
+/// - Parameter perform: The event handler to invoke if the event can be processed immediately.
+/// - Parameter enqueue: The event handler to invoke in the future if the event cannot currently be processed.
+typealias OnSinkEvent = (
+    _ perform: () -> Void,
+    _ enqueue: @escaping () -> Void
```
nabs-m: I wonder if there's a better word to convey `event handler to invoke in the future` if the event needs to be enqueued. Might be a nit pick, but `perform` was immediately clear to me, while for `enqueue` I initially thought this could be what runs when the event is enqueued (or maybe how you enqueue it) as opposed to a future action that happens because you need to enqueue this first. I'll throw in some ideas to help convey what I'm thinking, but let me know what you think: `postEnqueuePerform`, `enqueuedPerform`, `futurePerform`, `futureAction`, etc. I didn't use AI so there's probably better examples lol.

nabs-m (approved):
I don't have much experience with Workflow, so some of these concepts are new to me, but thank you for the clear explanation in the description!

Workflow/Sources/WorkflowHost.swift
jamieQ: tbh i forget exactly what i was thinking when i added this... 😅. i should probably just add a unit test that it starts in the expected state when initialized. (and really i should add tests for the various expected transitions)

- [x] TODO: improve the tests

Workflow/Sources/WorkflowHost.swift
jamieQ: how about `immediatePerform` vs `deferredPerform` or, replace 'perform' with some other word like 'handler' or something?

Workflow/Sources/SubtreeManager.swift
jamieQ: i agree, and think leaving something like this (at least for now[^1]) makes sense. i think we do still need to change the logic that gets enqueued – the way the existing code works fails in the 'someone spun the run loop before you finished a render pass' case, because those manual run loop turns can run that enqueued block, but it doesn't recurse into the `ReusableSink` method itself and re-check the validation state; it just unconditionally forwards through to the event pipe and we hope for the best (which seems to often crash in that edge case).

there's also an API design question here in my mind – who is responsible for making the check? i'm inclined to also move that logic out to the new `SinkEventHandler` type so the decision making is basically all in one place, but we'll need to pass the node-local state through to do that. it's a little awkward b/c the validation state enum is generic over various things, but we could just pass a `isPending` flag in the callback i suppose.

[^1]: it's conceivable to me that much of the node-local state could probably be moved out to the 'tree level', but haven't thought of a compelling reason to work though that at the moment

Workflow/Sources/WorkflowHost.swift
nabs-m: Yeah what you suggested is actually very clear that they're both actions to be performed, with one being right away and the other being later. The docs clarify that it's gonna be one or the other (based on whether it needs to be enqueued or not). I think `perform` is fine!

Workflow/Sources/SubtreeManager.swift
jamieQ: i made a couple changes to address this:
1. added a new method `withEventHandlingSuspended` (better names welcome) that can be used by the workflow host to explicitly ensure no synchronous event handlers will be run when updating the root node of the tree
2. restored the original handling logic & made the new event handling code paths conditional on a runtime configuration, so they will be opt-in and we can enable it via a feature flag

Workflow/Tests/WorkflowHostTests.swift
```diff
@@ -58,32 +61,90 @@ final class WorkflowHostTests: XCTestCase {
 final class WorkflowHost_EventEmissionTests: XCTestCase {
     // Previous versions of Workflow would fatalError under this scenario
     func test_event_sent_to_invalidated_sink_during_action_handling() {
-        let root = Parent()
-        let host = WorkflowHost(workflow: root)
+        let host = WorkflowHost(workflow: Parent())
+        let (lifetime, token) = ReactiveSwift.Lifetime.make()
+        defer { _ = token }
```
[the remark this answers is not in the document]
jamieQ: token needs to exist until end of scope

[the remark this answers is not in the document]
mjohnson12: AI usually gets reactive lifetimes wrong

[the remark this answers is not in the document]
jamieQ: we all have to learn it in our own way 😄

PR #382, opened 2025-09-19
Fix warnings in Xcode 26

jszumski:
Fix Xcode 26 warnings in sources copied from ComposableArchitecture that use SwiftSyntax (fixed in https://github.com/pointfreeco/swift-composable-architecture/pull/3160) and add dependencies to `IssueReporting` in modules that import it in tests to fix:

```
Package.swift: dependency 'xctest-dynamic-overlay' is not used by any target
```

## Checklist

- [ ] Unit Tests
- [ ] UI Tests


WorkflowSwiftUIMacros/Sources/Derived/Extensions.swift
robmaceachern, line 1: It looks like the changes here were related to resolving SwiftSyntax 510 / Xcode 15.3 warnings: https://github.com/pointfreeco/swift-composable-architecture/pull/2909/files#diff-4dd6925c27d6c7c053ed08b23987135c180ceb1239e35ba4f7e9d27538affc80

robmaceachern (approved):
I think the diff will be cleaned up after running the linter but changes lgtm

mjohnson12 (approved):

PR #383, opened 2025-09-22
[chore]: minor Package.swift cleanup

jamieQ:
update the `Package.swift` file to consistently use the `.product()` dependency case for external dependencies.


jamieQ:
this will just be rolled into other PRs

PR #389, opened 2026-04-09
Add native SwiftUI.Bindable support for iOS 17+

johnnewman-square:
This adds support for the `.sending(...)` API on iOS 17+ Bindables. The original implementation worked with `Perception.Bindable`, which is marked as obsolete on 17+, requiring clients to switch to `SwiftUI.Bindable` where these APIs didn't previously exist.

## Summary
- Adds `_NativeStoreBindable`, a native `@Bindable` equivalent of `_StoreBindable` for iOS 17+ clients, supporting all three `.sending()` overloads (`sink:action:`, `closure:`, and `action:` for `SingleActionModel`)
- Suppresses false-positive Perception runtime warnings on iOS 17+ by setting `skipPerceptionChecking` in `Store`'s state-reading and child-store scoping paths, where the `Perceptible` overload of `PerceptionRegistrar.access` is selected at compile time even though native observation is tracking
- Adds four native binding tests in `StoreTests` mirroring the existing `Perception.Bindable` tests, using `XCTSkip` to skip on iOS 16
- Adds `NativeMultiCounterView` sample demonstrating `@Bindable` usage without `WithPerceptionTracking`, wired via `#available` in `MultiCounterScreen`

## Test plan
- [x] Run `WorkflowSwiftUI-Tests` on an iOS 17+ simulator — verify the 4 native binding tests pass
- [x] Run `WorkflowSwiftUI-Tests` on an iOS 16 simulator — verify the native binding tests are skipped via `XCTSkip` and existing tests still pass
- [x] Build and run the `ObservableComposition` sample on iOS 17+ — verify `NativeMultiCounterView` is used with no Perception runtime warnings
- [x] Build and run the `ObservableComposition` sample on iOS 16 — verify `MultiCounterView` is used as fallback

🤖 Generated with [Claude Code](https://claude.com/claude-code)


johnnewman-square:
@codex review

WorkflowSwiftUI/Sources/Bindable+Store.swift
```diff
@@ -14,6 +14,68 @@ extension Perception.Bindable {
...
+    public subscript<Model: ObservableModel, Member>(
+        dynamicMember keyPath: KeyPath<Model.State, Member>
+    ) -> _NativeStoreBindable<Model, Member>
+        where Value == Store<Model>
+    {
+        _NativeStoreBindable(bindable: self, keyPath: keyPath)
+    }
+}
+
+/// Native `@Bindable` equivalent of `_StoreBindable` for iOS 17+.
```
robmaceachern, line 32: suggestion: Please add the iOS 17 `@Bindable` happy path to the adoption guide. This PR introduces a public native binding surface via `SwiftUI.Bindable`/`_NativeStoreBindable`, but the guide still only shows `@Perception.Bindable` inside `WithPerceptionTracking`, so it is hard to discover when callers can switch to the native wrapper and when they can drop `WithPerceptionTracking`. A short example in the bindings section would make the intended migration path much clearer.

---
*This response was drafted with AI assistance.*

WorkflowSwiftUI/Tests/StoreTests.swift
```diff
@@ -772,6 +772,145 @@ final class StoreTests: XCTestCase {
         await fulfillment(of: [countDidChange], timeout: 0)
         XCTAssertEqual(state.count, 1)
     }
+
+    // MARK: - Native SwiftUI Bindings
```
robmaceachern, line 776: suggestion: Please add a native child-store observation test here as well. The new iOS 17 suppression in `Store.access(...)` covers optional and collection child-store wrapper reads, but these added native tests only exercise scalar bindings through `store.count` and the `.sending(...)` overloads. Since the PR rationale explicitly includes false-positive warnings during child-store scoping, a regression in that path would still bypass CI. Even a small `store.optional` or `store.array`/`store.identified` observation test would close that gap.

---
*This response was drafted with AI assistance.*

robmaceachern (approved):

WorkflowSwiftUI/Tests/StoreTests.swift
johnnewman-square: That was a great idea, thanks. I've pushed up a new observation test.

WorkflowSwiftUI/Sources/Bindable+Store.swift
johnnewman-square: Done! The adoption guide has been updated.

johnnewman-square:
@robmaceachern I have addressed the feedback, thanks again. One thing I wanted to be sure to mention is that the `WithPerceptionTracking` warning logs will now be suppressed on iOS 17+ with these changes. They will still log on iOS 16.

PR #392, opened 2026-07-15
chore: Migrate core Workflow module off of ReactiveSwift

amorde:
## Changes

Remove `ReactiveSwift` usage from the core `Workflow` and `WorkflowUI` modules.

The only usages there were for emitting outputs from WorkflowHosts to external consumers. The internal implementation of Workflow itself did not use `ReactiveSwift` at all.

`WorkflowHost` API changes:
```diff
- public let rendering: Property<WorkflowType.Rendering>
+ public var rendering: WorkflowType.Rendering { get }

+ public var renderingPublisher: AnyPublisher<WorkflowType.Rendering, Never>

- public var output: Signal<WorkflowType.Output, Never> { get }
+ public var outputPublisher: AnyPublisher<WorkflowType.Output, Never> { get }
```

`WorkflowHostingController` API changes:
```diff
- public var output: Signal<Output, Never> { get }
+ public var outputPublisher: AnyPublisher<Output, Never> { get }
```

## Checklist

- [x] Unit Tests
- [ ] UI Tests

### DIFF
--- a/Workflow/Sources/WorkflowHost.swift
+++ b/Workflow/Sources/WorkflowHost.swift
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
+import Combine
 import Dispatch
-import ReactiveSwift
 
 /// Defines a type that receives debug information about a running workflow hierarchy.
 public protocol WorkflowDebugger {
@@ -33,16 +33,25 @@ public protocol WorkflowDebugger {
 
 /// Manages an active workflow hierarchy.
 public final class WorkflowHost<WorkflowType: Workflow> {
-    private let (outputEvent, outputEventObserver) = Signal<WorkflowType.Output, Never>.pipe()
+    private let outputSubject = PassthroughSubject<WorkflowType.Output, Never>()
 
     // @testable
     let rootNode: WorkflowNode<WorkflowType>
 
-    private let mutableRendering: MutableProperty<WorkflowType.Rendering>
+    private let renderingSubject: CurrentValueSubject<WorkflowType.Rendering, Never>
 
-    /// Represents the `Rendering` produced by the root workflow in the hierarchy. New `Rendering` values are produced
+    /// The current `Rendering` produced by the root workflow in the hierarchy. A new `Rendering` value is produced
     /// as state transitions occur within the hierarchy.
-    public let rendering: Property<WorkflowType.Rendering>
+    public var rendering: WorkflowType.Rendering {
+        renderingSubject.value
+    }
+
+    /// A publisher of the `Rendering` values produced by the root workflow in the hierarchy. Emits the current
+    /// `Rendering` when subscribed to, followed by a new value after each subsequent render pass. Use
+    /// `dropFirst()` to observe only changes.
+    public var renderingPublisher: AnyPublisher<WorkflowType.Rendering, Never> {
+        renderingSubject.eraseToAnyPublisher()
+    }
 
     /// Context object to pass down to descendant nodes in the tree.
     let context: HostContext
@@ -88,8 +97,7 @@ public final class WorkflowHost<WorkflowType: Workflow> {
             parentSession: nil
         )
 
-        self.mutableRendering = MutableProperty(rootNode.render())
-        self.rendering = Property(mutableRendering)
+        self.renderingSubject = CurrentValueSubject(rootNode.render())
         rootNode.enableEvents()
 
         debugger?.didEnterInitialState(snapshot: rootNode.makeDebugSnapshot())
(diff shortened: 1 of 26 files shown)


Workflow/Sources/WorkflowHost.swift
```diff
@@ -33,16 +33,24 @@ public protocol WorkflowDebugger {
...
     /// as state transitions occur within the hierarchy.
-    public let rendering: Property<WorkflowType.Rendering>
+    public var rendering: WorkflowType.Rendering {
+        renderingSubject.value
+    }
+
+    /// A publisher of the `Rendering` values produced by the root workflow in the hierarchy. Emits a new value
+    /// after each render pass that occurs following subscription. Use `rendering` to read the current value.
+    public var renderingPublisher: AnyPublisher<WorkflowType.Rendering, Never> {
+        renderingSubject.dropFirst().eraseToAnyPublisher()
```
johnnewman-square: 🤖 `renderingPublisher` applying `dropFirst()` at the public boundary removes the atomic current-value-plus-future-updates behavior previously available from `Property.producer`. A render between reading `rendering` and subscribing becomes the subject’s current value and is then dropped. Could this return `renderingSubject.eraseToAnyPublisher()` and move `.dropFirst()` to `WorkflowHostingController`, which already consumes the initial rendering separately?

Workflow/Sources/WorkflowHost.swift
```diff
@@ -33,16 +33,24 @@ public protocol WorkflowDebugger {
 
 /// Manages an active workflow hierarchy.
 public final class WorkflowHost<WorkflowType: Workflow> {
-    private let (outputEvent, outputEventObserver) = Signal<WorkflowType.Output, Never>.pipe()
+    private let outputSubject = PassthroughSubject<WorkflowType.Output, Never>()
```
johnnewman-square, line 36: 🤖 Could `WorkflowHost.deinit` send `.finished` to both subjects, with lifecycle tests? The old `MutableProperty` completed and the output pipe terminated when their owner deinitialized, whereas a Combine subscription retains its subject. Without explicit completion, retained sinks and `publisher.values` loops can remain alive after the host is released.

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
```diff
@@ -26,8 +26,8 @@ public final class WorkflowHostingController<ScreenType, Output>: WorkflowUIView
     public typealias CustomizeEnvironment = (inout ViewEnvironment) -> Void
 
     /// Emits output events from the bound workflow.
-    public var output: Signal<Output, Never> {
-        workflowHost.output
+    public var output: AnyPublisher<Output, Never> {
```
johnnewman-square: 🤖 The PR description promises `WorkflowHostingController.outputPublisher`, and `WorkflowHost` follows that naming, but this API remains `output` while changing its type. Could we rename it and its preview/test call sites to `outputPublisher`, or update the migration contract if retaining `output` is intentional?

johnnewman-square (commented):
This looks good on my end. My agent just had a few suggestions.

Workflow/Sources/WorkflowHost.swift
amorde: This was done because previously `rendering.signal` produced only changed renderings so without this all existing `rendering.signal` callers had to include `.dropFirst()`, which felt like a footgun for migrations. 

Because ReactiveSwift's `Property` provides _both_ options, we have to pick which behavior we adopt here:
* `Property.signal`: Only emits updates
* `Property.producer`: Emits current value + updates

Workflow/Sources/WorkflowHost.swift
amorde: Fixed

WorkflowUI/Sources/Hosting/WorkflowHostingController.swift
amorde: Fixed - went with `outputPublisher`

Workflow/Sources/WorkflowHost.swift
amorde: Keeping the `producer` semantics and letting callers use `.dropFirst()` feels more idiomatic (for example, matches the behavior of `@Published` properties), so I'll go with that.

johnnewman-square: Thanks Eric! Matching the idiomatic `@Published` behavior feels like the best approach to me, too.

johnnewman-square (approved):
Thanks for making those tweaks.

PR #393, opened 2026-07-16
Make Perception warning suppression opt-in

johnnewman-square:
PRs #389 and #390 automatically suppressed Perception runtime warnings on platforms using native Observation. This restores the warnings by default while preserving an opt-in suppression path.

## Summary

- Add `Runtime.Configuration.suppressPerceptionCheckingWhenUsingObservation`, defaulting to `false`
- Continue normal Store access unless suppression is explicitly enabled
- Apply suppression only in debug builds on iOS 17+, macOS 14+, tvOS 17+, and watchOS 10+
- Cover both direct state reads and child Store wrapper access
- Add tests for the default warning behavior and opt-in suppression

## Test plan

- [x] Run `StoreTests` on iOS 17.5 — 25 passed
- [x] Run `StoreTests` on iOS 16.2 — passed with 8 expected skips
- [x] Run the ObservableComposition sample app on iOS 17 to ensure the warnings are occurring by default.

## Checklist

- [x] Unit Tests
- [x] UI Tests (not applicable)
- [x] Snapshot Tests (not applicable)

### DIFF
--- a/Workflow/Sources/RuntimeConfiguration.swift
+++ b/Workflow/Sources/RuntimeConfiguration.swift
@@ -43,7 +43,8 @@ public enum Runtime {
 
     private static var _defaultConfiguration = Configuration()
 
-    static var configuration: Configuration {
+    /// The configuration active for the current task, falling back to the default configuration.
+    public static var configuration: Configuration {
         _currentConfiguration ?? _defaultConfiguration
     }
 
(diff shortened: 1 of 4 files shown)


Workflow/Sources/RuntimeConfiguration.swift
```diff
@@ -43,7 +43,8 @@ public enum Runtime {
 
     private static var _defaultConfiguration = Configuration()
 
-    static var configuration: Configuration {
+    /// The configuration active for the current task, falling back to the default configuration.
+    public static var configuration: Configuration {
         _currentConfiguration ?? _defaultConfiguration
     }
```
johnnewman-square, line 49: - [x] Investigate switching to `package` instead of `public`.

amorde (approved):

Workflow/Sources/RuntimeConfiguration.swift
johnnewman-square: Moved to `package`.

